PR #70, opened 2015-01-08
CDK-865,  CDK-846 : New topics based on old kite-data-module-overview.

DennisDawson:


introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in HDFS.
+
+The default storage _scheme_ (as opposed to a _[schema](#schemas)_, described below) is Hive. If you omit the scheme in your URI, your dataset is stored in Hive. For example, the URI `movies` is equivalent to `dataset:hive:/movies`. A dataset you create with this URI is stored as a table in Hive's `default` database.
```
rdblue: There isn't a default scheme in the API, so we shouldn't mention one in the overview. This is how the CLI behaves because we used to identify datasets by name instead of by the whole URI. We kept support for dataset names so old invocations would still work, which means that it seems like there is a default URI pattern.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+Dataset URIs cannot contain periods (.).
```
rdblue: Dataset names and namespaces can't contain periods for compatibility with Avro and Hive. Other parts of the URI can.

It would be better to state that we put a restriction on the name and namespace parts of the URI, which are used when the underlying system (like Hive) supports them. Names and namespaces must be made of alphanumeric characters or underscore and can't start with a number. I think record field names have the same requirement.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+title: Introduction to Datasets
+---
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in HDFS.
```
rdblue: I think it would be valuable to say that URIs provide a few things:
1. The underlying storage (which you already cover here) based on the scheme
2. A name and a namespace. These are used when the underlying system has the same concept (as with Hive) and can also be listed using the [Datasets.list method](http://kitesdk.org/docs/0.17.1/apidocs/org/kitesdk/data/Datasets.html#list(java.lang.String)).
3. Sometimes, a location that could be a HDFS path

This would help later when you note that the namespace and name components have restrictions on them.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+The default storage _scheme_ (as opposed to a _[schema](#schemas)_, described below) is Hive. If you omit the scheme in your URI, your dataset is stored in Hive. For example, the URI `movies` is equivalent to `dataset:hive:/movies`. A dataset you create with this URI is stored as a table in Hive's `default` database.
+
+Dataset URIs cannot contain periods (.).
+
+You can create datasets in Hive, HDFS, HBase, or as local files. See [Dataset and View URIs]({{site.baseurl}}/URIs.html).
```
rdblue: Could you add more information about what "Dataset and View URIs" contains? Like "See [Dataset and View URIs](...) for the available URI patterns and more detailed options."

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+```json
+{
+  "type":"record",
+  "name":"Movie",
+  "namespace":"org.kitesdk.examples.data",
+  "fields":[
+    {"name":"id","type":"int"},
+    {"name":"title","type":"string"},
+    {"name":"releaseDate","type":"string"},
+    {"name":"imdbUrl","type":"string"}
```
rdblue: Best practice is to use snake_case rather than camelCase for field names because there are reports of inconsistent handling of upper case letters. So this should be "release_date" and "imdb_url".

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+| [Using the schema of an Avro data file][api-schema-from-data] | [Inferring a schema from a CSV file][cli-schema-from-csv] |
+
+[api-schema-from-class]: {{site.baseurl}}/Inferring-a-Schema-from-a-Java-Class.html
+[api-schema-from-data]: {{site.baseurl}}/Inferring-a-Schema-from-an-Avro-Data-File.html
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Configuration Options
+
+While you need only a URI and schema to create a dataset, you can enhance the performance of your dataset with additional configuration.
```
rdblue: I think partitioning should be a top-level section rather than being part of Configuration Options. The best practice is to always consider how the data should be partitioned. A partition strategy is optional because sometimes the best partition strategy is to not partition, not because it can be safely ignored.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Partitions
+
+Partitions define logical categories for data storage. For example, you might most often retrieve your data using time-based queries. You can define a partitioning strategy by year, month, day, and hour. When you look for data from January 8, 2015 between 8:00 and 9:00 a.m., your search engine only has to look in the data partition `/2015/1/8/8`. By using partitions that correspond to your most common queries, your data searches run more quickly.
+
+You define your partition strategy in [JSON format][json-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategy].
```
rdblue: This should link to the docs for creating a partition strategy with the CLI and the docs for the builder API, like we do with Schemas above.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Configuration Options
+
+While you need only a URI and schema to create a dataset, you can enhance the performance of your dataset with additional configuration.
+
+### Partitions
+
+Partitions define logical categories for data storage. For example, you might most often retrieve your data using time-based queries. You can define a partitioning strategy by year, month, day, and hour. When you look for data from January 8, 2015 between 8:00 and 9:00 a.m., your search engine only has to look in the data partition `/2015/1/8/8`. By using partitions that correspond to your most common queries, your data searches run more quickly.
```
rdblue: We always need to be careful with time examples to avoid confusion about how time zones are handled. The time we use for the partitioning is _always_ in UTC. Your example will be correct if the local time zone lines up with UTC. Otherwise, looking for data between 8 and 9 (local time) usually looks in partition /2015/1/8/16 to account for the zone. I suggest we avoid this entirely by removing the hour part of the example.

rdblue: The actual partition would be `/year=2015/month=1/day=8/hour=8`.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+[json-format]: {{site.baseurl}}/Partition-Strategy-Format.html
+[partition-strategy]: {{site.baseurl}}/Partitioned-Datasets.html
+
+
+### Column Mapping
+
+Column mapping allows you to configure how your records should be stored in HBase for maximum performance and efficiency. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. See [Column Mapping][column-mapping].
+
+[column-mapping]: {{site.baseurl}}/Column-Mapping.html
```
rdblue: This should also link to the CLI command (mapping-config) and the builder API.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+While you need only a URI and schema to create a dataset, you can enhance the performance of your dataset with additional configuration.
+
+### Partitions
+
+Partitions define logical categories for data storage. For example, you might most often retrieve your data using time-based queries. You can define a partitioning strategy by year, month, day, and hour. When you look for data from January 8, 2015 between 8:00 and 9:00 a.m., your search engine only has to look in the data partition `/2015/1/8/8`. By using partitions that correspond to your most common queries, your data searches run more quickly.
+
+You define your partition strategy in [JSON format][json-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategy].
+
+[json-format]: {{site.baseurl}}/Partition-Strategy-Format.html
+[partition-strategy]: {{site.baseurl}}/Partitioned-Datasets.html
```
rdblue: The Schema section has an example. Maybe that would be helpful here? Something like:

```
user@work:~$ kite-dataset partition-config ts:year ts:month ts:day -s rating.avsc
[ {
  "name" : "year",
  "source" : "ts",
  "type" : "year"
}, {
  "name" : "month",
  "source" : "ts",
  "type" : "month"
}, {
  "name" : "day",
  "source" : "ts",
  "type" : "day"
} ]
```

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Column Mapping
+
+Column mapping allows you to configure how your records should be stored in HBase for maximum performance and efficiency. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. See [Column Mapping][column-mapping].
+
+[column-mapping]: {{site.baseurl}}/Column-Mapping.html
+
+### Properties
+
+You can add custom settings to dataset properties at creation time using the `--set` option. For example, you might create the dataset _users_ using a cache size of 20 (rather than the default cache size of 10):
```
rdblue: This is specific to the CLI. The main point is to say that you can store configuration properties on your dataset that are relevant for reading and writing the data. `kite.writer.cache-size` is a good example, but we should make it clear that this is using the CLI to set a property. This also gives us a good place to link to docs on more properties and tuning.

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+kite-dataset create users --schema user.avsc --set kite.writer.cache-size=20
+```
+
+You can also define key-value pairs to use as custom properties in your application by appending the option `--set prop.name=value`.
```
rdblue: This would come before the example that uses it, right?

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+Column mapping allows you to configure how your records should be stored in HBase for maximum performance and efficiency. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. See [Column Mapping][column-mapping].
+
+[column-mapping]: {{site.baseurl}}/Column-Mapping.html
+
+### Properties
+
+You can add custom settings to dataset properties at creation time using the `--set` option. For example, you might create the dataset _users_ using a cache size of 20 (rather than the default cache size of 10):
+
+```
+kite-dataset create users --schema user.avsc --set kite.writer.cache-size=20
```
rdblue: We should also link to the relevant [property API on DatasetDescriptor](http://kitesdk.org/docs/0.17.1/apidocs/org/kitesdk/data/DatasetDescriptor.Builder.html#property(java.lang.String,%20java.lang.String)).

introduction-to-datasets.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+You can also define key-value pairs to use as custom properties in your application by appending the option `--set prop.name=value`.
+
+## Working with Datasets
```
rdblue: I think this is more of a "next steps" section once people have the context that we intend to provide with this page. Next steps are to go through the tutorial with CSV data or view the data with the CLI, Hive, Impala, etc.

introduction-to-datasets.md
DennisDawson: I added the description before the link, but as a rule I always want to say "See blank" with no additional text, because it's obvious we're saying "go there for more information."

introduction-to-datasets.md
DennisDawson: Do you mean that we should link to a page like this one that talks about CDH tuning in general?

http://www.cloudera.com/content/cloudera/en/documentation/core/latest/topics/cdh_admin_performance.html

I found this in the POM

  <properties>
    
    <kite.version>0.16.0</kite.version>
    <kite.avro.version>1.7.5-cdh5.1.0</kite.avro.version>
    <kite.crunch.version>0.10.0-cdh5.1.0</kite.crunch.version>
    <kite.findbugs.version>2.0.1</kite.findbugs.version>
    <kite.hive.version>0.12.0-cdh5.1.0</kite.hive.version>
    <kite.oozie.version>4.0.0-cdh5.1.0</kite.oozie.version>
    <kite.spark.version>1.0.0-cdh5.1.0</kite.spark.version>

    
    <hadoop.guava.version>11.0.2</hadoop.guava.version>
    <hadoop.log4j.version>1.2.16</hadoop.log4j.version>
    <hadoop.slf4j.version>1.6.1</hadoop.slf4j.version>
  </properties>

Is there an exhaustive list of the possible settings someplace? It would probably be a good idea to have a reference page that lists at least the most popular properties and suggestions for why a user might want to change them.

rdblue: > Do you mean that we should link to a page like this one that talks about CDH tuning in general?

I'm referring to CDK-846.

For POM properties, let's not worry about them for now. I'm talking about descriptor properties here.

introduction-to-datasets.md
DennisDawson: Need to talk about this --the property API doesn't really give any more information. Need to document properties and what to do with them.

rdblue: Yes: CDK-846

introduction-to-datasets.md
rdblue: Not just stored in HDFS: stored in the /user/cloudera/datasets/movies directory in HDFS.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in HDFS.
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). You can also retrieve the values using the `Datasets.list` method.
```
rdblue: You can list the available datasets (both namespace and name) using the `Datasets.list` method.

Also, the method reference should link to the javadoc for it.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+Often, the name and namespace provide a location that could be an HDFS path.
```
rdblue: I'm not sure what you're trying to say here. The name and namespace are parts of the path, but do not form a path on their own.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+To ensure compatibility with underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (_) and cannot start with a number.
```
rdblue: Referring to "underlying systems" sounds like magic. Could you replace that with something more specific, like "to ensure compatibility with Hive or other underlying systems..."?

rdblue: Also, underscores that aren't markdown formatting should be escaped: \\\_ => \_

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+You can create datasets in Hive, HDFS, HBase, or as local files. There are several URI patterns available and additional options.See [Dataset and View URIs]({{site.baseurl}}/URIs.html).
```
rdblue: Missing space between sentences.

Also, I would prefer longer URLs (like URIs.html) as [reference links](http://kramdown.gettalong.org/syntax.html#reference-links) to make it easier to read and maintain the raw markup.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+  "fields":[
+    {"name":"id","type":"int"},
+    {"name":"title","type":"string"},
+    {"name":"release_date","type":"string"},
+    {"name":"imdb_url","type":"string"}
+  ]
+}
+```
+
+The goal is to get the schema into `.avsc` format. The following links provide examples for inferring schemas from data files or Java classes.
```
rdblue: Why is this a goal?

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+Often, the name and namespace provide a location that could be an HDFS path.
+
+To ensure compatibility with underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (_) and cannot start with a number.
+
+You can create datasets in Hive, HDFS, HBase, or as local files. There are several URI patterns available and additional options.See [Dataset and View URIs]({{site.baseurl}}/URIs.html).
+
+## Schemas
+
+A schema defines the field names and datatypes for a dataset. Kite relies on an Apache Avro schema definition for each dataset. For example, this is the schema definition for a table using four columns from the `movies` dataset.[<sup>1</sup>](#notes)
```
rdblue: Why does Kite use Avro schemas?

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+| [Using the schema of an Avro data file][api-schema-from-data] | [Inferring a schema from a CSV file][cli-schema-from-csv] |
+
+[api-schema-from-class]: {{site.baseurl}}/Inferring-a-Schema-from-a-Java-Class.html
+[api-schema-from-data]: {{site.baseurl}}/Inferring-a-Schema-from-an-Avro-Data-File.html
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Partitions
+
+Partitions define logical categories for data storage. For example, you might most often retrieve your data using time-based queries. You can define a partitioning strategy by year, month, day, and hour. When you look for data from January 8, 2015, your search engine only has to look in the data partition `/year=2015/month=1/day=8`. By using partitions that correspond to your most common queries, your data searches run more quickly.
```
rdblue: A search engine is a very specific application and probably isn't what you want to use here. Similarly, search isn't the appropriate verb in the last sentence.

We don't need to include hour in the partition strategy if the example uses days.

While you can think of partitions as logical categories, I don't think that's the best way to get the idea across because the point isn't just to categorize, it is to structure data in a way that allows you to ignore most of it, most of the time.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+Partitioning is optional, because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+When using the CLI, you define your partition strategy in [JSON format][json-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategy].
```
rdblue: The CLI helps build partition strategies with the partition-config command (should link to it). The JSON format is the canonical representation and is what that command produces, as well as what the other commands (like create) expect. The JSON format is also used by the API, so be careful that the wording isn't misleading.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+When using the API, you define your dataset in the `DatasetDescriptor.Builder`. See [Partition Builder][partition-builder].
```
rdblue: This is one way to produce a partition strategy, but you can also pass the JSON representation as a String or as a file reference. Maybe this should just link to the API and note that the JSON representation is what you will primarily use when working with the CLI, as I noted above.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+For example, you can use the CLI command `partition-config` to define a partition strategy using name:value pairs to specify that the dataset should be partitioned on a timestamp (`ts`) field by year, month, and day.
+
+```
+user@work:~$ kite-dataset partition-config ts:year ts:month ts:day -s rating.avsc
+```
+The result is a partition definition in JSON format. You can use the `-o` option to out put the result and store it in a .json file.
```
rdblue: Missing a line between the code block and the next paragraph.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+} ]
+```
+
+[json-format]: {{site.baseurl}}/Partition-Strategy-Format.html
+[partition-strategy]: {{site.baseurl}}/Partitioned-Datasets.html#partition-strategies
+[partition-builder]: {{site.baseurl}}/API-Overview.html#partition-strategy
+
+## Configuration Options
+
+While you need only a URI and schema to create a dataset, you can enhance the performance of your dataset with additional configuration.
```
rdblue: You can use properties for additional control of how Kite reads or writes your data, which can be used for performance tuning.

rdblue: I just realized this is to introduce the column mapping and properties sections, so my comment above isn't quite accurate. The note about enhancing the performance isn't correct. As I noted above, properties do more than just performance and are intended to expose some control over how Kite behaves while reading or writing. Column mappings affect performance, but the real take-away is that column mappings are required to store data in HBase.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Column Mapping
+
+Column mapping allows you to configure how your records should be stored in HBase for maximum performance and efficiency. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly.
```
rdblue: This has an effect on performance, but to state "form maximum performance..." is misleading. HBase datasets require column mappings. Column mappings have an effect on performance.

introduction-to-datasets.md
```diff
@@ -0,0 +1,119 @@
...
+Column mapping allows you to configure how your records should be stored in HBase for maximum performance and efficiency. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. 
+
+See [Column Mapping][column-mapping]. See also the [CLI mapping-config command][cli-mapping-config].
+
+[column-mapping]: {{site.baseurl}}/Column-Mapping.html
+[cli-mapping-config]: {{site.baseurl}}/cli-reference.html#mapping-config
+
+### Properties
+
+You can add custom settings to dataset properties at creation time using the `--set` option. For example, in the CLI you might create the dataset _users_ using a cache size of 20 (rather than the default cache size of 10):
```
rdblue: The part about the `--set` option is CLI-specific and should be part of the example, but not part of the paragraph explaining what properties do.

This should also briefly explain what the writer cache size controls, which is the number of files that are kept open at one time. One file is used for each partition that you write to.

using-kite-properties.md
```diff
@@ -0,0 +1,32 @@
...
+title: Using Kite Properties
+---
+
+Kite gives you the option of changing settings of existing system properties to enhance performance. You also have the option of creating custom properties.
+
+## Setting Kite Properties
+
+There are several dataset properties that affect how files are written. These settings are especially relevant for Parquet because it buffers records in memory. 
+
+`kite.write.cache-size` controls the number of files kept open by an HDFS or Hive dataset writer.
```
rdblue: Could you restructure so both properties here are separate sections? I think that would be more clear.

Also, I'd make a sub-section of the parquet.block.size section called "Avoiding Parquet OutOfMemoryExceptions".

introduction-to-datasets.md
DennisDawson: I don't know. Because it's a standard? Spelunking about, I only see references that say it's a good idea to use them, but no reason why. (https://github.com/kite-sdk/kite-examples/tree/master/json)

tomwhite: One reason is that Avro is the only binary format that is supported across all the components in Hadoop.

rdblue: We chose to standardize on Avro for a few reasons:
* The format is widely used, supported across components (as Tom noted), and mature
* Avro supports all 3 data models we're interested in exposing (generic, reflect, specific)
* Avro Schemas work with both Avro and Parquet, the main data formats we're interested in
* Avro Schemas are expressive enough and we don't hear too many complaints about lack of some structures or concepts

That's obviously way too much information to put here, but I'd like to see some justification rather than just stating that it is the case.

using-kite-properties.md
DennisDawson: I don't mind adding headings, and I've done so. Generally, it's non-standard to create subheadings when there is only one. The reason I put the two properties together is that they act in concert, and they are the only two properties we're documenting, at the moment. This will settle out when we add descriptions of additional properties.

using-kite-properties.md
```diff
@@ -0,0 +1,37 @@
...
+
+### parquet.block.size
+
+Kite passes descriptor properties to the underlying file formats. For example, Parquet defines `parquet.block.size`, which is approximately the amount of data that is buffered before writing a group of records (a "row group"). `parquet.block.size` defaults to 128MB.
+
+#### Avoiding Parquet OutOfMemory Exceptions
+The amount of data kept in memory for each file could be up to the Parquet block size in bytes. That means that the upper bound for a writer's memory consumption is `parquet.block.size` * `kite.writer.cache-size`. It is important that this number doesn't exceed a reasonable portion of the heap memory allocated to the process, or else the write could fail with an `OutOfMemoryException`. You can set these properties on your datasets with [`DatasetDescriptor.Builder.property`][dataset-descriptor-builder] when using the API, or with the `--set` option from the command line.
```
rdblue: Missing an empty line after the heading

rdblue: Shouldn't the note about how to set properties be elsewhere?

using-kite-properties.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
+```
+
+Note that Cloudera does not recommend decreasing the `parquet.block.size`.
+
+## Creating Custom Properties
+
+In addition to setting existing system properties, you can create your own key-value pairs to use as custom properties in your application. When using the  [CLI create command]({{site.baseurl}}/cli-reference.html#create), you add custom properties with the option `--set prop.name=value`.
```
rdblue: You can also use `--set` with update to add them to a dataset.

using-kite-properties.md
DennisDawson: Moved the "how to" above kite.write.cache-size.

using-kite-properties.md
DennisDawson: Added update option and linked to CLI.

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in HDFS.
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). You can retrieve the names and namespaces for the available datasets using the [`Datasets.list`](http://kitesdk.org/docs/current/apidocs/org/kitesdk/data/Datasets.html#list(java.net.URI)) method.
```
rdblue: This is another place where a reference link would be better.

Also the javadoc for `Datasets.list`:
> List the Dataset URIs in the repository identified by the URI

While you can see the names and namespaces, it would be more valuable to call out that the method gives URIs for valid datasets underneath the repository URI. If we are mentioning a repository URI, then we should add it to the URIs document.

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+To ensure compatibility with Hive and other underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (\_) and cannot start with a number.
+
+You can create datasets in Hive, HDFS, HBase, or as local files. There are several URI patterns available and additional options. See [Dataset and View URIs]({{site.baseurl}}/URIs.html).
+
+## Schemas
+
+A schema defines the field names and datatypes for a dataset. Kite relies on an Apache Avro schema definition for each dataset. The Kite team made this choice because Avro supports all three data models exposed in Kite (generic, reflect, specific) and also works with Parquet format.
```
rdblue: How about "Kite standardizes data definition by using Avro Schemas for both Parquet and Avro, and supports all 3 standard Avro object models (generic, specific, and reflect)."

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in HDFS.
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). You can retrieve the names and namespaces for the available datasets using the [`Datasets.list`](http://kitesdk.org/docs/current/apidocs/org/kitesdk/data/Datasets.html#list(java.net.URI)) method.
+
+To ensure compatibility with Hive and other underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (\_) and cannot start with a number.
+
+You can create datasets in Hive, HDFS, HBase, or as local files. There are several URI patterns available and additional options. See [Dataset and View URIs]({{site.baseurl}}/URIs.html).
```
rdblue: This should also be a reference link.

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Partitions
+
+Partitions define logical divisions for data storage. For example, you might most often work with data using time-based queries. You can define a partitioning strategy by year, month, and day. When you are using data from January 8, 2015, Hadoop only has to access data stored in the partition `/year=2015/month=1/day=8`. By using partitions that correspond to your most common queries, your applications run more quickly.
+
+Partitioning is optional, because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+You define your partition strategy in [JSON format][json-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategy].
+
+When using the API, you define your dataset in the [`DatasetDescriptor.Builder`](http://kitesdk.org/docs/current/apidocs/org/kitesdk/data/DatasetDescriptor.Builder.html#partitionStrategy(java.io.File)).
```
rdblue: Should be a reference link.

using-kite-properties.md
```diff
@@ -0,0 +1,40 @@
...
+
+Note that Cloudera does not recommend decreasing the `parquet.block.size`.
+
+## Creating Custom Properties
+
+In addition to setting existing system properties, you can create your own key-value pairs to use as custom properties in your application. When using the  CLI [`create`]({{site.baseurl}}/cli-reference.html#create) or [`update`]({{site.baseurl}}/cli-reference.html#update)command, you add custom properties with the option `--set prop.name=value`.
+
+When using the Kite API, you can add properties using [DatasetDescriptor.Builder.property][dataset-descriptor-builder].
+
+[dataset-descriptor-builder]: http://kitesdk.org/docs/current/apidocs/org/kitesdk/data/DatasetDescriptor.Builder.html#property(java.lang.String,%20java.lang.String).
```
rdblue: This should also use `{{site.baseurl}}` instead of the full URL.

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+You can add custom settings to dataset properties at creation time.
+
+For example, Kite's default cache size is 10. This is the number of files that are open at any one time; each file represents a partition in the dataset written to by your application. In the CLI, you can use the `--set` option to increase the cache size of 20:
+
+```
+kite-dataset create users --schema user.avsc --set kite.writer.cache-size=20
+```
+
+You can also create custom properties for your own applications. See [Using Kite Properties]({{site.baseurl}}/using-kite-properties.html).
```
rdblue: Should be a reference link.

using-kite-properties.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
+```
+
+Note that Cloudera does not recommend decreasing the `parquet.block.size`.
+
+## Creating Custom Properties
+
+In addition to setting existing system properties, you can create your own key-value pairs to use as custom properties in your application. When using the  CLI [`create`]({{site.baseurl}}/cli-reference.html#create) or [`update`]({{site.baseurl}}/cli-reference.html#update)command, you add custom properties with the option `--set prop.name=value`.
```
rdblue: Should be reference links

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Next Steps
+
+The example [Using the Kite Command Line Interface to Create a Dataset]({{site.baseurl}}/Using-the-Kite-CLI-to-Create-a-Dataset.html) demonstrates many of the features described above, and shows you how to create your own Hadoop dataset using only three commands.
```
rdblue: Should be a reference link

introduction-to-datasets.md
```diff
@@ -0,0 +1,123 @@
...
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Partitions
+
+Partitions define logical divisions for data storage. For example, you might most often work with data using time-based queries. You can define a partitioning strategy by year, month, and day. When you are using data from January 8, 2015, Hadoop only has to access data stored in the partition `/year=2015/month=1/day=8`. By using partitions that correspond to your most common queries, your applications run more quickly.
+
+Partitioning is optional, because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+You define your partition strategy in [JSON format][json-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategy].
```
rdblue: "JSON format" refers to any JSON, so I think this should be more clear and link to the partition strategy format, which is JSON.

introduction-to-datasets.md
tomwhite: Avro Reflect is not supported under Parquet, so we should leave the original wording, or explicitly point it out.

rdblue: Good point, let's just list generic and specific for now.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+While you need only a URI and schema to create a dataset, you can use properties for additional control over how Kite reads or writes your data.
+
+### Column Mapping
+
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. 
+
+See [Column Mapping][cm]. See also the [CLI mapping-config command][climc].
+
+[cm]: {{site.baseurl}}/Column-Mapping.html
```
rdblue: It looks like you've gone a little too far to the other extreme. "mapping" and "cli-mapping" or similar would be fine here. We want it short (not the full title) but not to the point where it loses meaning.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/datasets/movies` is stored in the `/user/cloudera/datasets/movies` directory in HDFS.
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive).
+
+To ensure compatibility with Hive and other underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (\_) and cannot start with a number.
```
rdblue: This talks about the restrictions but doesn't identify which parts of the URI are used as the name and namespace. It is confusing that certain, unidentified parts have restrictions. I think you just need to add that this is typically the last two components of the URI path. Also point out in the example above the namespace and the dataset name and give a Hive example.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+You can create datasets in Hive, HDFS, HBase, or as local files. There are several URI patterns available and additional options. See [Dataset and View URIs][uris].
+
+[list]: {{site.baseurl}}/apidocs/org/kitesdk/data/Datasets.html#list(java.net.URI)
+[uris]:{{site.baseurl}}/URIs.html
+
+## Schemas
+
+A schema defines the field names and datatypes for a dataset. Kite relies on an Apache Avro schema definition for each dataset. Kite standardizes data definition by using Avro schemas for both Parquet and Avro, and supports the standard Avro object models _generic_ and _specific_.
```
rdblue: I thought the "Kite standardizes data definition..." sentence was to replace "Kite relies on..."

Do we need to state both things separately? Isn't it clear from the second that Kite uses Avro Schemas for its data definitions?

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Partitions
+
+Partitions define logical divisions for data storage. For example, you might most often work with data using time-based queries. You can define a partitioning strategy by year, month, and day. When you are using data from January 8, 2015, Hadoop only has to access data stored in the partition `/year=2015/month=1/day=8`. By using partitions that correspond to your most common queries, your applications run more quickly.
+
+Partitioning is optional, because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+You define your partition strategy in [partition strategy JSON format][psjf] and apply it when you create your dataset. See [Partitioned Datasets][ps].
+
+When using the API, you define your dataset in the [`DatasetDescriptor.Builder`][ddb].
```
rdblue: This sentence doesn't make sense on its own. What are you trying to get across here?

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+When using the API, you define your dataset in the [`DatasetDescriptor.Builder`][ddb]. 
+
+For example, you can use the CLI command [`partition-config`][clipc] to define a partition strategy using name:value pairs to specify that the dataset should be partitioned on a timestamp (`ts`) field by year, month, and day.
+
+```
+user@work:~$ kite-dataset partition-config ts:year ts:month ts:day -s rating.avsc
+```
+
+The result is a partition definition in JSON format. You can use the `-o` option to out put the result and store it in a .json file.
```
rdblue: The output option isn't needed here, though it is fine if you think it is going to be helpful and answer a question that the reader will have at this point.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+Partitioning is optional, because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+You define your partition strategy in [partition strategy JSON format][psjf] and apply it when you create your dataset. See [Partitioned Datasets][ps].
+
+When using the API, you define your dataset in the [`DatasetDescriptor.Builder`][ddb]. 
+
+For example, you can use the CLI command [`partition-config`][clipc] to define a partition strategy using name:value pairs to specify that the dataset should be partitioned on a timestamp (`ts`) field by year, month, and day.
+
+```
+user@work:~$ kite-dataset partition-config ts:year ts:month ts:day -s rating.avsc
```
rdblue: I think it would be valuable to explain why the schema is passed to this command

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+[ddb]:{{site.baseurl}}/apidocs/org/kitesdk/data/DatasetDescriptor.Builder.html#partitionStrategy(java.io.File)
+[ps]: {{site.baseurl}}/Partitioned-Datasets.html#partition-strategies
+[clipc]: {{site.baseurl}}/cli-reference.html#partition-config
+[psjf]: {{site.baseurl}}/Partition-Strategy-Format.html
+
+## Configuration Options
+
+While you need only a URI and schema to create a dataset, you can use properties for additional control over how Kite reads or writes your data.
```
rdblue: This mentions properties specifically but not the column mapping in the next section. This should be cleaned up to represent the entire section.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Column Mapping
+
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. 
+
+See [Column Mapping][cm]. See also the [CLI mapping-config command][climc].
```
rdblue: It seems awkward to use "See X. Also see Y". Why not "See X and Y"?

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+[clipc]: {{site.baseurl}}/cli-reference.html#partition-config
+[psjf]: {{site.baseurl}}/Partition-Strategy-Format.html
+
+## Configuration Options
+
+While you need only a URI and schema to create a dataset, you can use properties for additional control over how Kite reads or writes your data.
+
+### Column Mapping
+
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly.
```
rdblue: What does "data-centric" mean?

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. 
+
+See [Column Mapping][cm]. See also the [CLI mapping-config command][climc].
+
+[cm]: {{site.baseurl}}/Column-Mapping.html
+[climc]: {{site.baseurl}}/cli-reference.html#mapping-config
+
+### Properties
+
+You can add custom settings to dataset properties at creation time.
```
rdblue: The note about using properties for additional control should be here (or in both places) to explain what properties do, rather than just stating that you can add them. The first question I have when I read this sentence is "Why would I do that?"

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+
+For example, Kite's default cache size is 10. This is the number of files that are open at any one time; each file represents a partition in the dataset written to by your application. In the CLI, you can use the `--set` option to increase the cache size of 20:
```
rdblue: This is good to help explain, but I still think that the purpose of properties should be more clear.

introduction-to-datasets.md
```diff
@@ -0,0 +1,130 @@
...
+kite-dataset create users --schema user.avsc --set kite.writer.cache-size=20
+```
+
+You can also create custom properties for your own applications. See [Using Kite Properties][useprop].
+
+[useprop]:{{site.baseurl}}/using-kite-properties.html
+
+## Next Steps
+
+The example [Using the Kite Command Line Interface to Create a Dataset][usecli] demonstrates many of the features described above, and shows you how to create your own Hadoop dataset using only three commands.
```
rdblue: This doesn't need the full title. Why not use something shorter, like "The Kite CLI example demonstrates..."?

using-kite-properties.md
```diff
@@ -0,0 +1,42 @@
...
+
+There are several dataset properties that affect how files are written. These settings are especially relevant for Parquet because it buffers records in memory. 
+
+You can set properties on your datasets with [`DatasetDescriptor.Builder.property`][dataset-descriptor-builder] when using the API, or with the `--set` option from the command line.
+
+### kite.write.cache-size
+
+`kite.write.cache-size` controls the number of files kept open by an HDFS or Hive dataset writer.
+
+Writers open one file per partition to which they write records. When the writer receives a record that goes in a new partition (one for which there isn't an open file) it creates a new file in that partition. If the number of open files exceeds the cache size, Kite closes the file that was used least recently.
```
rdblue: I would like to have an example of when you would want to increase this setting. A good one is when you're writing a year worth of data to year/month partitions. There are 12 partitions and 10 open files, so you'll be constantly closing and opening files, which slows down your writes. It is better to increase the writer cache to 12 so that you can have all 12 months open at once.

using-kite-properties.md
```diff
@@ -0,0 +1,42 @@
...
+
+#### Avoiding Parquet OutOfMemory Exceptions
+
+The amount of data kept in memory for each file could be up to the Parquet block size in bytes. That means that the upper bound for a writer's memory consumption is `parquet.block.size` * `kite.writer.cache-size`. It is important that this number doesn't exceed a reasonable portion of the heap memory allocated to the process, or else the write could fail with an `OutOfMemoryException`. 
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
+```
+
+Note that Cloudera does not recommend decreasing the `parquet.block.size`.
```
rdblue: Did I say this? I don't think this is entirely correct, but I'm wondering what I originally said that sounded like this so I can think through what I meant to say.

using-kite-properties.md
```diff
@@ -0,0 +1,42 @@
...
+### parquet.block.size
+
+Kite passes descriptor properties to the underlying file formats. For example, Parquet defines `parquet.block.size`, which is approximately the amount of data that is buffered before writing a group of records (a "row group"). `parquet.block.size` defaults to 128MB.
+
+#### Avoiding Parquet OutOfMemory Exceptions
+
+The amount of data kept in memory for each file could be up to the Parquet block size in bytes. That means that the upper bound for a writer's memory consumption is `parquet.block.size` * `kite.writer.cache-size`. It is important that this number doesn't exceed a reasonable portion of the heap memory allocated to the process, or else the write could fail with an `OutOfMemoryException`. 
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
```
rdblue: It is good to say that you should only keep 2 files open at once, but we need to cover how to make this happen without constantly opening and closing files. The trick is to organize your data so that you only need to write to one file at once. That sounds hard, but we provide Crunch tools to do it: [`CrunchDatasets.partition`](http://kitesdk.org/docs/0.17.1/apidocs/org/kitesdk/data/crunch/CrunchDatasets.html#partition(org.apache.crunch.PCollection,%20org.kitesdk.data.View)) will partition a collection to be stored efficiently in a dataset. That means it will structure data to write only one file at a time.

using-kite-properties.md
DennisDawson: https://issues.cloudera.org/browse/CDK-846

"Users can set the two properties on their datasets, though we don't recommend decreasing parquet.block.size."

introduction-to-datasets.md
DennisDawson: When linking to another topic, I use the full title to avoid confusion. The exception is in the TOC, where the titles are too long for the format we're using.

introduction-to-datasets.md
DennisDawson: Expanded the link names so that they are short but readable, ~10-15 chars.

using-kite-properties.md
rdblue: Got it. What I meant is that it is recommended that you avoid out of memory problems by decreasing the number of writers rather than decreasing the parquet block size. So I'm not saying that you shouldn't go lower than 128MB.

Sorry I wasn't clear. The parquet.block.size (row group size) setting should be configured based on other concerns, and you should structure your program to avoid writing to more files at one time than you have available memory to handle.

introduction-to-datasets.md
DennisDawson: The first sentence says that Avro is used to define dataset schemas. The second says that Avro is used for both Parquet and Avro, which was one of the advantages listed in response to the question "Why does Kite use Avro?" If we just say that Kite uses Avro schemas for both Parquet and Avro, that could be interpreted that other schemas might be used for other storage types. It doesn't hurt to be explicit.

introduction-to-datasets.md
DennisDawson: Yup, that was an orphan. I moved it to the top and paired it with a sentence saying that you use the `create` command in the CLI.

introduction-to-datasets.md
DennisDawson: I moved the sentence after the JSON output. To me, it doesn't make sense to run the command in isolation: we either need to give an example where the user would copy and paste the output, or suggest they save it to a file.

introduction-to-datasets.md
DennisDawson: When you want me to "explain why," please provide the explanation. This saves me the trouble of guessing at what you want me to say. I'm guessing that you pass the schema to provide the definitions for whatever fields are being partitioned.

introduction-to-datasets.md
DennisDawson: "While you need only a URI and schema to create a dataset, you can use column mapping and properties for additional control over how Kite reads or writes your data. "

If we add another section under configuration options in the future, I'll come back and try to think of a generic description.

introduction-to-datasets.md
DennisDawson: The rationale is that They should go see Column Mapping for more info, and they might also be interested in the CLI command I don't want to imply that the reader needs to go see both. It might be awkward, but it's better than a sentence that says "See column mapping for more information about column mapping."

introduction-to-datasets.md
DennisDawson: I pulled this paragraph directly from the Column Mapping topic. I think the topic came from Adam. I understand it to mean that column mapping strategy is inferred from the data values rather than programmatic concerns. We can remove this term, or if you have something more descriptive, please make a suggestion.

introduction-to-datasets.md
DennisDawson: I added the statement, but as you say, it already appears in the section description. The sentence also leads directly into an example, which shows one reason you might add a property. It doesn't hurt to be redundant, though, because readers might skim or skip down to the Properties heading, so I don't object to redundancy. Redundancy is okay.

introduction-to-datasets.md
rdblue: That makes sense. Maybe the problem is that there's not context for why you might follow either link. That's why I'm generally uncomfortable with the "See X" link style anyway. What about adding description? For example, "to use the CLI to build a mapping config JSON file for your Schema, see mapping-config command."

DennisDawson: That sort of description before a link is an outdated standard. I think this is similar to your preference that links open in the same window: you have to give the reader some credit.

 It's clear that the difference is that Column Mapping is a general description, while CLI mapping config refers to the command line interface.

introduction-to-datasets.md
DennisDawson: Yup, added the description.

introduction-to-datasets.md
rdblue: I can see how it isn't strictly necessary for the column mapping link. It's clear that the link goes to more description on column mappings. I'm not sure that it is clear from the mapping-config command that the command helps you generate column mapping JSON files. There shouldn't be a link without the reader having enough information to choose whether or not to click it.

introduction-to-datasets.md
rdblue: That's fine.

using-kite-properties.md
DennisDawson: Added:

For example, you might have an application writing a year's worth of data to year/month partitions. There are 12 partitions, but only 10 files open at the same time. The application must constantly close and open files, which slows down your writes. If you increase the writer cache to 12, you can have files for all 12 months open at once.

using-kite-properties.md
DennisDawson: What other concerns...? Or can I omit that part, and just say that developers should structure their program...etc.?

introduction-to-datasets.md
rdblue: The partition strategy is validated using the Schema. This catches problems early. For example, if you use "timestmap" then this will catch the typo because the Schema has a "timestamp" column. It also validates that field types can be used. You can't extract the year from a string, for example.

introduction-to-datasets.md
rdblue: Then should it be "all" datasets?

using-kite-properties.md
rdblue: Yes, we don't need to discuss how or why you might change the Parquet block size. Just say that the recommended way to avoid OOM is to write to fewer files and that the `CrunchDatasets.partition` methods are intended to help with that.

introduction-to-datasets.md
DennisDawson: Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/animals/donkeys` is stored in the `/user/cloudera/animals/donkeys` directory in HDFS, with _animals_ as the namespace and _donkeys_ as the dataset name.

URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). The name and namespace are typically the last two values in a URI. For example, the URI `dataset:hive:animals/donkeys` creates a dataset named _donkeys_ in the _animals_ namespace.

introduction-to-datasets.md
DennisDawson: Yes, that's better.

using-kite-properties.md
DennisDawson: The recommended way to avoid out of memory errors is to write to fewer files. To help with that, [CrunchDatasets.partition][cd-partition] methods restructure the parallel collection so that all of the entities stored in a given partition are processed by the same writer.

using-kite-properties.md
DennisDawson: This should be addressed by the addition, above.

using-kite-properties.md
rdblue: Sounds good, other than that we need to use "exception" instead of "error" because Error is a thing in Java and we don't want to be confusing.

DennisDawson: D'oh! I hate it when I do that. Corrected to "The recommended way to avoid out of memory exceptions..."

introduction-to-datasets.md
DennisDawson: Added:

"Kite validates your partition strategy using the dataset schema. This catches problems early. For example, if you mistakenly type in your partition definition _timestmap_, validation catches the typo because the schema has a _timestamp_ column. Kite also validates that field types can be partitioned on the selected value. For example, you can't extract a year from a string."

using-kite-properties.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Setting Kite Properties
+
+There are several dataset properties that affect how files are written. These settings are especially relevant for Parquet because it buffers records in memory. 
+
+You can set properties on your datasets with [`DatasetDescriptor.Builder.property`][dataset-descriptor-builder] when using the API, or with the `--set` option from the command line.
+
+### kite.write.cache-size
+
+`kite.write.cache-size` controls the number of files kept open by an HDFS or Hive dataset writer.
```
DennisDawson: Added "The default cache size is 10."

using-kite-properties.md
```diff
@@ -0,0 +1,46 @@
...
+
+#### Avoiding Parquet OutOfMemory Exceptions
+
+The amount of data kept in memory for each file could be up to the Parquet block size in bytes. That means that the upper bound for a writer's memory consumption is `parquet.block.size` * `kite.writer.cache-size`. It is important that this number doesn't exceed a reasonable portion of the heap memory allocated to the process, or else the write could fail with an `OutOfMemoryException`. 
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
+```
+
+The recommended way to avoid out of memory exceptions is to write to fewer files. To help with that, [CrunchDatasets.partition][cd-partition] methods restructure the parallel collection so that all of the entities stored in a given partition are processed by the same writer.
```
DennisDawson: I don't think the context is unclear. I'll change "parallel collection" to "collected data" and change "entities" to "records." 

"The recommended way to avoid out of memory exceptions is to write to fewer files. To help with that, [CrunchDatasets.partition][cd-partition] methods restructure the collected data so that all of the records stored in a given partition are processed by the same writer."

The important point is that all of the items for one partition are processed by the same writer, and I think that's clear.

The original language comes from the Javadoc for this item. The term "parallel collection" is used in all four variations of the partition method. If all I have to go by is the Javadoc, and the Javadoc is misleading or lacks context, it makes for a very difficult writing process.

I am not the source of information. During the markup process, if reviewers feel that the descriptions are incorrect, they need to provide the correct information, rather than saying it needs "context." I will handle the style once I know what point you're trying to make.

introduction-to-datasets.md
```diff
@@ -0,0 +1,136 @@
+---
+layout: page
+title: Introduction to Datasets
+---
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
+
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+When using the CLI, you define your dataset using the [`create`][create] command. When using the API, you define your dataset using [`DatasetDescriptor.Builder`][dataset-desc-bldr].
```
DennisDawson: The context of this line changed when I moved it here and added the note about the CLI. It had been in the section on properties. I've changed the link to the `Datasets.create` command, which will lead to `DatasetDescriptor`, etc.

using-kite-properties.md
```diff
@@ -0,0 +1,46 @@
...
+
+Kite gives you the option of changing settings of existing system properties to enhance performance. You also have the option of creating custom properties for additional control over how Kite reads or writes your data.
+
+## Setting Kite Properties
+
+There are several dataset properties that affect how files are written. These settings are especially relevant for Parquet because it buffers records in memory. 
+
+You can set properties on your datasets with [`DatasetDescriptor.Builder.property`][dataset-descriptor-builder] when using the API, or with the `--set` option from the command line.
+
+### kite.write.cache-size
```
rdblue: This is actually `kite.writer.cache-size`. Looks like there was a typo in the original issue.

using-kite-properties.md
```diff
@@ -0,0 +1,46 @@
...
+### parquet.block.size
+
+Kite passes descriptor properties to the underlying file formats. For example, Parquet defines `parquet.block.size`, which is approximately the amount of data that is buffered before writing a group of records (a "row group"). `parquet.block.size` defaults to 128MB.
+
+#### Avoiding Parquet OutOfMemory Exceptions
+
+The amount of data kept in memory for each file could be up to the Parquet block size in bytes. That means that the upper bound for a writer's memory consumption is `parquet.block.size` multiplied by the `kite.writer.cache-size`. It is important that this number doesn't exceed a reasonable portion of the heap memory allocated to the process, or else the write could fail with an `OutOfMemoryException`. 
+
+```
+kite-dataset update <uri> --set kite.writer.cache-size=2
```
rdblue: This command needs more context.

using-kite-properties.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+The recommended way to avoid out of memory exceptions is to write to fewer files. To help with that, [CrunchDatasets.partition][cd-partition] methods restructure the collected data so that all of the records stored in a given partition are processed by the same writer.
```
rdblue: This still doesn't quite work. Just "restructure data so that all of the records..." should be fine. You also need to note that this is limited to Crunch.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
+---
+layout: page
+title: Introduction to Datasets
+---
+
+A dataset is a collection of records, similar to a relational database table. Records are similar to table rows, but the columns can contain not only strings or numbers, but also nested data structures such as lists, maps, and other records.
```
rdblue: What about changing this to "records are like table rows"? Using "similar to" twice sounds odd.

rdblue: I think you should also rephrase to get rid of the two "but" phrases in the second sentence.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+[create]:{{site.baseurl}}/cli-reference.html#create
+[datasets-create]:{{site.baseurl}}/apidocs/org/kitesdk/data/Datasets.html#create(java.net.URI,%20org.kitesdk.data.DatasetDescriptor)
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/animals/donkeys` is stored in the `/user/cloudera/animals/donkeys` directory in HDFS.
+
+### Names and Namespaces 
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). The name and namespace are typically the last two values in a URI. For example, if you create a dataset using the URI `dataset:hive:animals/donkeys`, Kite stores a Hive dataset named _donkeys_ in the _animals_ namespace. If you create a dataset using the URI `dataset:hdfs:/user/cloudera/animals/donkeys`, Kite stores an HDFS dataset named _donkeys_ in the _animals_ namespace.
```
rdblue: I think the "Kite stores a Hive dataset named..." sentence should use Hive's terminology rather than Kite's. So the URI `dataset:hive:animals/donkeys` will be stored in the Hive table _donkeys_ in the _animals_ Hive db.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+To define a dataset, Kite minimally requires a [URI](#uris) and a [schema](#schemas).
+
+When using the CLI, you define your dataset using the [`create`][create] command. When using the API, you define your dataset using the [Datasets.create][datasets-create] method.
+
+[create]:{{site.baseurl}}/cli-reference.html#create
+[datasets-create]:{{site.baseurl}}/apidocs/org/kitesdk/data/Datasets.html#create(java.net.URI,%20org.kitesdk.data.DatasetDescriptor)
+
+## URIs
+
+Kite identifies datasets by URI. The URI you provide tells Kite how and where to store data. For example, a dataset created with the URI `dataset:hdfs:/user/cloudera/animals/donkeys` is stored in the `/user/cloudera/animals/donkeys` directory in HDFS.
```
rdblue: Can we come up with a more meaningful example than "animals/donkeys"? How about "fact_tables/ratings"?

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Names and Namespaces 
+
+URIs also define a name and namespace for your dataset. Kite uses these values when the underlying system has the same concept (for example, Hive). The name and namespace are typically the last two values in a URI. For example, if you create a dataset using the URI `dataset:hive:animals/donkeys`, Kite stores a Hive dataset named _donkeys_ in the _animals_ namespace. If you create a dataset using the URI `dataset:hdfs:/user/cloudera/animals/donkeys`, Kite stores an HDFS dataset named _donkeys_ in the _animals_ namespace.
+
+To ensure compatibility with Hive and other underlying systems, names and namespaces in URIs must be mode of alphanumeric characters or underscore (\_) and cannot start with a number.
+
+### URI Schemes
+
+You can create datasets using the schemes _Hive_, _HDFS_, _HBase_, or as local files. There are several URI patterns available and additional options. See [Dataset and View URIs][uris].
```
rdblue: Schemes are case-sensitive so these should be lower case. But, it looks like you're actually trying to list what the storage options are, judging by the last entry, "or as local files". I think this should either list the actual schemes or just note that you can store data in Hive, HDFS, HBase, or the local file system.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+| [Using the schema of an Avro data file][api-schema-from-data] | [Inferring a schema from a CSV file][cli-schema-from-csv] |
+
+[api-schema-from-class]: {{site.baseurl}}/Inferring-a-Schema-from-a-Java-Class.html
+[api-schema-from-data]: {{site.baseurl}}/Inferring-a-Schema-from-an-Avro-Data-File.html
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Partitions
+
+Partitions define optional logical divisions for data storage. For example, you might most often work with data using time-based queries. You can define a partitioning strategy by year, month, and day. When you are using data from January 8, 2015, Hadoop only has to access data stored in the partition `/year=2015/month=1/day=8`. By using partitions that correspond to your most common queries, your applications run more quickly.
```
rdblue: Reminder: we talked about changing "optional" here.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+
+Partitioning is optional because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
```
rdblue: Reminder: this should be inverted: You should always consider how to partition your datasets. However, there are times when partitioning data is inefficient or unnecessary, so supplying a partition strategy is optional.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+| --------                                                      | ---------------------- |
+| [Inferring a schema from a Java Class][api-schema-from-class] | [Inferring a schema from a Java class][cli-schema-from-class] |
+| [Using the schema of an Avro data file][api-schema-from-data] | [Inferring a schema from a CSV file][cli-schema-from-csv] |
+
+[api-schema-from-class]: {{site.baseurl}}/Inferring-a-Schema-from-a-Java-Class.html
+[api-schema-from-data]: {{site.baseurl}}/Inferring-a-Schema-from-an-Avro-Data-File.html
+[cli-schema-from-class]: {{site.baseurl}}/cli-reference.html#obj-schema
+[cli-schema-from-csv]: {{site.baseurl}}/cli-reference.html#csv-schema
+
+## Partitions
```
rdblue: The other section headings refer to things that you supply to Kite: Schemas and URIs. This doesn't fit, so I think it should be changed to "Partition Strategies".

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+
+Partitioning is optional because there are times when partitioning is not the most efficient solution. However, you should always consider partitioning as a best practice when planning your dataset.
+
+Kite validates your partition strategy using the dataset schema. This catches problems early. For example, if you mistakenly type in your partition definition _timestmap_, validation catches the typo because the schema has a _timestamp_ column. Kite also validates that field types can be partitioned on the selected value. For example, you can't extract a year from a string.
+
+You define your partition strategy in [partition strategy JSON format][ps-format] and apply it when you create your dataset. See [Partitioned Datasets][partition-strategies].
+
+For example, you can use the CLI command [`partition-config`][cli-part-conf] to define a partition strategy using name:value pairs to specify that the dataset should be partitioned on a timestamp (`ts`) field by year, month, and day.
```
rdblue: The previous sections talks about supplying a partition strategy when you create a dataset. The "for example" connector makes me think that the example is going to cover creating a dataset, but this instead talks about using the CLI to define the format.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+} ]
+```
+
+If you want to save the partition definition for use with the CLI or API, you can use the `-o` option to output the result to a `.json` file.
+
+[partition-strategies]: {{site.baseurl}}/Partitioned-Datasets.html#partition-strategies
+[cli-part-conf]: {{site.baseurl}}/cli-reference.html#partition-config
+[ps-format]: {{site.baseurl}}/Partition-Strategy-Format.html
+
+## Configuration Options
```
rdblue: Along the same lines as the comment on the previous section heading, I think this should be "Additional Configuration".

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+
+While you need only a URI and schema to create a dataset, you can use column mapping and properties for additional control over how Kite reads or writes your data.
+
+### Column Mapping
+
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly.
```
rdblue: This should read "Kite uses a column mapping to configure..." instead of "HBase requires...". HBase doesn't require the mapping, Kite does to store data in HBase.

rdblue: I also think it would be better to give an example or clarify further what "data-centric" means.

introduction-to-datasets.md
```diff
@@ -0,0 +1,142 @@
...
+HBase requires that you use column mapping to configure how your records should be stored. You define column mapping in JSON format in a data-centric way. Kite stores and retrieves the data correctly. 
+
+See [Column Mapping][column-mapping]. See also the [CLI mapping-config command][cli-column-mapping].
+
+[column-mapping]: {{site.baseurl}}/Column-Mapping.html
+[cli-column-mapping]: {{site.baseurl}}/cli-reference.html#mapping-config
+
+### Properties
+
+You can add custom settings to dataset properties at creation time for additional control over how Kite reads or writes your data.
```
rdblue: This mixes the two uses of properties that are outlined in the properties doc: custom settings and additional control over how Kite reads and writes. The initial paragraph should be "You can add properties to the dataset to control how Kite reads and writes your data" and the follow-on paragraph can address custom properties (that one looks fine as-is).

using-kite-properties.md
DennisDawson: Changed this in the header and the next sentence. Filed CDK-901 to be sure this is fixed in the update -h output.

using-kite-properties.md
DennisDawson: Moved it up under the example for kite.writer.cache-size and changed it to match.

using-kite-properties.md
DennisDawson: Changed to "The recommended way to avoid out of memory exceptions is to write to fewer files. When working with Crunch, you can use [CrunchDatasets.partition][cd-partition] methods to restructure data so that all of the records stored in a given partition are processed by the same writer."

rdblue: Sounds great.

introduction-to-datasets.md
DennisDawson: _Like_ can be interpreted to mean _identical_. _Similar to_, while redundant, is more clear.

_But_ the conjunction is not the same as _not only...but also_, a correlative conjunction; this is fine grammatically and conveys the proper meaning.

introduction-to-datasets.md
DennisDawson: Changed to "For example, if you create a dataset using the URI `dataset:hive:animals/donkeys`, Kite stores a Hive table _donkeys_ in the _animals_ Hive database." [db is used in syntax, but I checked with the Hive writer who showed me that it's spelled out in docs.]

introduction-to-datasets.md
DennisDawson: The hierarchy is intuitively clear with _animals/donkeys_. I don't see how fact_tables/ratings is more meaningful.

introduction-to-datasets.md
DennisDawson: Changed to:

"You can create datasets in Hive, HDFS, HBase, or as local files. The dataset schemes are defined using scheme-specific URI patterns. See [Dataset and View URIs][uris]."

introduction-to-datasets.md
rdblue: Perhaps meaningful isn't a great way to put it. How about useful? Something closer to how it would be used.

introduction-to-datasets.md
DennisDawson: Removed the word _optional_.

introduction-to-datasets.md
DennisDawson: Changed to:

"For example, you can create a partition strategy using the CLI command [`partition-config`][cli-part-conf] with name:value pairs to specify that the dataset should be partitioned on a timestamp (`ts`) field by year, month, and day."

introduction-to-datasets.md
DennisDawson: Changed heading to "Additional Configuration"

introduction-to-datasets.md
DennisDawson: Data-centric is both intuitive and an industry term. http://en.wikipedia.org/wiki/Data-centric_programming_language 

I don't want to give an example, here, because it would require as much explanation as the link provides.

I think the point here is that other than column mapping, Kite interacts with HBase the way it does with the other schemes.

"Kite uses a column mapping to configure how your records should be stored. You define column mapping in JSON format. Kite stores and retrieves HBase data seamlessly, using the same commands as with other storage schemes."

If that doesn't work, please provide exactly what you think should be said, so that I don't have to guess at the point we're trying to make here.

introduction-to-datasets.md
DennisDawson: Okay, but I want to say "set" rather than "add" properties, then, to make the distinction.

"You can set properties at creation time for additional control over how Kite reads or writes your data."

introduction-to-datasets.md
rdblue: That is almost there, but the first sentence should indicate that this is for HBase

DennisDawson: "Kite uses a column mapping to configure how your records should be stored in HBase. You define column mapping in JSON format. Kite stores and retrieves HBase data seamlessly, using the same commands as with other storage schemes."

introduction-to-datasets.md
rdblue: "Set" is fine with me.

introduction-to-datasets.md
rdblue: That works.

introduction-to-datasets.md
rdblue: Here's a version that solves both problems. It uses _like_ and uses a single _but_ to clarify how they are not identical to rows: "Records are like table rows, but can contain lists, maps, and nested records, in addition to primitive types such as strings or numbers."

DennisDawson: After review by our editor, the sentence is fine and remains as written.

tomwhite:
I committed this change (in https://github.com/kite-sdk/kite-docs/commit/d78e28c84a1d6ae09f203e048eeb6e925073ea61). Thanks @DennisDawson!

PR #73, opened 2015-02-19
Slight cleanup of `awk` (avoiding UUOC)

BMDan:
Kill a UUOC, and `sort -n` in case we ever have any post-September 8, 2001 dates in the file (when Unix timestamp started to have an additional digit).


tomwhite:
Thanks @BMDan!

PR #75, opened 2015-02-23
CDK-29, CDK-466, CDK-827: Create Dataset, Load with Flume

DennisDawson:
The modules for preparing the VM and creating a dataset are modular, and can be used independently of the loading with Flume example.


tomwhite:
Overall this looks good. I think we need to think a bit more about the directory layout to avoid everything being in a flat hierarchy. How about putting everything under _tutorials_, then the Flume page under _tutorials/flume_, and the common material directly in _tutorials_.

DennisDawson:
Are you referring to kite-docs or kite-examples? I understand that new materials in kite-examples might go into a tutorial directory (for example, GenerateEvents.java), but we weren't going to fiddle with the existing examples. I'm happy with whatever we decide, but I don't want to have to redo/undo changes.

tomwhite:
I'm referring to kite-docs.

DennisDawson:
@rdblue Is there anything stopping us from going to a deeper directory structure, as Tom suggests?

rdblue:
@DennisDawson I just talked with Tom about this and there was some confusion. He thought that we were still using folders for individual pages, so the flume tutorial would be _tutorials/flume_. It should be _tutorials/flume.html_ instead. So the tutorials should all go in the _tutorials_ folder for now.

DennisDawson:
That's fine if there will only be one example (as is the case with Flume). If there is more than one example for a particular technology/feature, we'll consider adding another directory level.

preparing-the-vm.md
```diff
@@ -0,0 +1,91 @@
+---
+layout: page
+title: Preparing the Virtual Machine
```
rdblue: Shouldn't this be in the tutorials folder?

preparing-the-vm.md
```diff
@@ -0,0 +1,91 @@
...
+
+Complete the following steps to run Kite example code on a Cloudera Quickstart VM.
+
+* Install a VirtualBox or VMWare [Cloudera QuickStart VM][getvm] version 5.2 or later.
+* In that VM, run the following command from a terminal window. This command clones a local copy of Kite code examples you can build and run.
+
+```bash
+git clone https://github.com/kite-sdk/kite-examples.git
+```
+* Download the `kite-dataset` CLI JAR. To install the Kite CLI using `curl` and give execution permission to `kite-dataset.jar`, run the following commands from a terminal window.
```
rdblue: It would be better not to require the user to update the version of kite-dataset in the VM. Is there something not working that requires this update?

DennisDawson:
This PR has the files both at the root and in /tutorials. Moving to another PR.

PR #76, opened 2015-02-23
Per CDK-886, add instructions for replacing kite-dataset in /usr/bin

DennisDawson:
Instructions for replacing the installed version of kite-dataset with the latest version.

### DIFF
--- a/Install-Kite.md
+++ b/Install-Kite.md
@@ -11,3 +11,19 @@ chmod +x {{site.dataset-command}}
 ```
 
 The first command downloads the Kite tools JAR to a local file named _{{site.dataset-command}}_. The second changes the access rights on the JAR so that the classes in the JAR are executable.
+
+Some versions of the Cloudera QuickStart VM comes with Kite installed. However, it might not be the latest version. To check the version, from a terminal window, enter the following command.
+
+```
+kite-dataset --version
+```
+
+If the version number doesn't match the version you downloaded, you have two options.
+
+1. Include the path to `kite-dataset` when you invoke commands. For example, if the JAR file is in the current directory, use `./kite-dataset`.
+
+1. Replace the installed version of `kite-dataset.jar`.
+  a. Download `kite-dataset.jar` to the home directory.
+  b. In a terminal window on the VM, navigate to `/usr/bin`.
+  c. Enter the command `sudo cp ~/kite-dataset ./kite-dataset`
+  d. To verify that the correct version is in use, enter the command `kite-dataset --version`.


Install-Kite.md
```diff
@@ -11,3 +11,19 @@ chmod +x {{site.dataset-command}}
...
+
+```
+kite-dataset --version
+```
+
+If the version number doesn't match the version you downloaded, you have two options.
+
+1. Include the path to `kite-dataset` when you invoke commands. For example, if the JAR file is in the current directory, use `./kite-dataset`.
+
+1. Replace the installed version of `kite-dataset.jar`.
```
rdblue: No need to use the `jar` file extension, this is usually called `kite-dataset`. Also, this is `{{site.dataset-command}}` above, should that be used here as well?

Install-Kite.md
```diff
@@ -11,3 +11,19 @@ chmod +x {{site.dataset-command}}
...
+  a. Download `kite-dataset.jar` to the home directory.
```
rdblue: No need to specify a target directory.

Install-Kite.md
```diff
@@ -11,3 +11,19 @@ chmod +x {{site.dataset-command}}
...
+  b. In a terminal window on the VM, navigate to `/usr/bin`.
+  c. Enter the command `sudo cp ~/kite-dataset ./kite-dataset`
```
rdblue: Steps b and c should be combined into a single command.

This also replaces a wrapper script rather than the command, so we need to figure out how to replace the underlying file by having two sets of instructions: for packages and for parcels. The wrapper script sets up the java classpath so we don't want to overwrite it.

rdblue: For packages, replace /usr/lib/kite/bin/kite-dataset. For parcels, it is /opt/cloudera/parcels/<version>/lib/kite/bin/kite-dataset.

Install-Kite.md
DennisDawson: Replaced all occurrences with {{site.dataset-command}}

Install-Kite.md
DennisDawson: So, wait, is this a general CDH issue, and not something specific to the QuickStart VM? Packages/parcels would not be an issue on the VM, would it? Noting this above, as well.

Install-Kite.md
DennisDawson: Line 15 - Some versions of the Cloudera QuickStart VM [come] with Kite installed. Should this be Some versions of CDH come with Kite installed?

Install-Kite.md
DennisDawson: Okay, added more explanation. I don't want to use a variable name like <local directory>/ if I can avoid it. I just said "in this example, it is downloaded to the `~/` directory."

Install-Kite.md
```diff
@@ -10,4 +10,22 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
 chmod +x {{site.dataset-command}}
 ```
 
-The first command downloads the Kite tools JAR to a local file named _{{site.dataset-command}}_. The second changes the access rights on the JAR so that the classes in the JAR are executable.
+The first command downloads the Kite tools JAR to a local file named `{{site.dataset-command}}`. The second changes the access rights on the JAR so that the classes in the JAR are executable.
+
+Some versions of CDH come with `{{site.dataset-command}}` installed. However, it might not be the latest version. To check the version, from a terminal window, enter the following command.
```
rdblue: I think this should start a new section so readers have an obvious choice between installing or replacing an existing command.

Install-Kite.md
```diff
@@ -10,4 +10,22 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+
+```
+{{site.dataset-command}} --version
+```
+
+If the version number doesn't match the version you downloaded, you have two options.
```
rdblue: These instructions download a specific version, so you can just state the version here. There's a chance that the user is reading an old version of this guide, so it is also a good time to point them at the latest, which is always at /docs/current/Install-Kite.html

Install-Kite.md
```diff
@@ -10,4 +10,22 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+
+1. Include the path to `{{site.dataset-command}}` when you invoke commands. For example, if the JAR file is in the current directory, use `./{{site.dataset-command}}`.
+
+1. Replace the installed version of `{{site.dataset-command}}`.
+  a. Download `{{site.dataset-command}}` (in this example, it is downloaded to the `~/` directory).
+  b. Replace the old version of `{{site.dataset-command}}` in the existing location.
+    1. For packages, use `sudo cp ~/{{site.dataset-command}} /usr/bin/{{site.dataset-command}}`.
```
rdblue: For packages the path is `/usr/lib/kite/bin/{{site.dataset-command}}`

Install-Kite.md
```diff
@@ -10,4 +10,22 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+    1. For parcels, use `sudo cp ~/{{site.dataset-command}} /opt/cloudera/parcels//lib/kite/bin/{{site.dataset-command}}`.
+  
+    c. To verify that the correct version is in use, enter the command `{{site.dataset-command}} --version`.
```
rdblue: This is indented to with the sub-bullets under b. I'm not sure if it looks correct when rendered so you should check.

Install-Kite.md
DennisDawson: Working in the 5.2 VM, it seems that /usr/lib/kite/bin/ is not in the $PATH statement. Replacing that instance doesn't make a difference, but replacing the one in /usr/bin does. Not sure how you want to resolve that.

Install-Kite.md
DennisDawson: This renders correctly in Safari, which is why I did it that way. MD is not great at nested lists.

Install-Kite.md
rdblue: The file in /usr/bin initializes the java environment and then calls the one in /usr/lib/kite/bin. Replacing the one in the Kite install ensures that package updates work and that the java environment is correct. Otherwise there are more issues.

Install-Kite.md
rdblue: I don't really know if it does or not, but I wouldn't be surprised either way. Probably best to make sure it's fixed.

Install-Kite.md
DennisDawson: Yeah, okay. The reason I hesitate to do that is that I'm pretty sure everyone will always need to do something in order to use the latest version of the software. I'll add the heading "Verifying the Version In Use," which will hopefully get them to keep reading.

Install-Kite.md
DennisDawson: Ach - got it. I was working in a 5.2 VM where I'd changed the /usr/bin instance, which you informed me is only supposed to be a pointer to the one in `/usr/lib/kite/bin`.

DennisDawson: Question, though: is it a good practice to give the same name to the pointer JAR and the kite-dataset JAR? Could the one in /usr/bin be named `kite`, for example, to distinguish from the one with the goodies inside?

rdblue: They use the same name because they represent the same command. It's not really a problem and is a common pattern when relocating commands.

Install-Kite.md
```diff
@@ -3,6 +3,8 @@ layout: page
 title: Install the Kite Command Line Interface
 ---
 
+These are the instructions for downloading the Kite command line interface. The most up-to-date instructions for downloading the current version of the software are always available at [http://kitesdk.org/docs/current/Install-Kite.html](http://kitesdk.org/docs/current/Install-Kite.html).
```
rdblue: I think this should include the version that will be installed so that the "current" link makes a bit more sense.

Install-Kite.md
```diff
@@ -10,4 +12,24 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+```
+
+If the version number doesn't match the version you downloaded, you have two options.
+
+1. Include the path to `{{site.dataset-command}}` when you invoke commands. For example, if the JAR file is in the current directory, use `./{{site.dataset-command}}`.
+
+1. Replace the installed version of `{{site.dataset-command}}`.
+  a. Download `{{site.dataset-command}}` (in this example, it is downloaded to the `~/` directory).
+  b. Replace the old version of `{{site.dataset-command}}` in the existing location.
+    1. For packages, use `sudo cp ~/{{site.dataset-command}}/usr/lib/kite/bin/{{site.dataset-command}}`.
```
rdblue: This is missing a space between the two paths, as is the next option.

Install-Kite.md
```diff
@@ -10,4 +12,24 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+    1. For parcels, use `sudo cp ~/{{site.dataset-command}}/opt/cloudera/parcels/lib/kite/bin/{{site.dataset-command}}`.
+  
+    c. To verify that the correct version is in use, enter the command `{{site.dataset-command}} --version`.
```
rdblue: Nit: Indentation is off.

rdblue: This should state the expected version when the command is successful.

Install-Kite.md
```diff
@@ -10,4 +12,24 @@ curl http://central.maven.org/maven2/org/kitesdk/kite-tools/{{site.version}}/kit
...
+
+If the version number doesn't match the version you downloaded, you have two options.
+
+1. Include the path to `{{site.dataset-command}}` when you invoke commands. For example, if the JAR file is in the current directory, use `./{{site.dataset-command}}`.
+
+1. Replace the installed version of `{{site.dataset-command}}`.
+  a. Download `{{site.dataset-command}}` (in this example, it is downloaded to the `~/` directory).
+  b. Replace the old version of `{{site.dataset-command}}` in the existing location.
+    1. For packages, use `sudo cp ~/{{site.dataset-command}}/usr/lib/kite/bin/{{site.dataset-command}}`.
+    1. For parcels, use `sudo cp ~/{{site.dataset-command}}/opt/cloudera/parcels/lib/kite/bin/{{site.dataset-command}}`.
```
rdblue: The target should be `/opt/cloudera/parcels/CDH/lib/kite/bin/kite-dataset` The CDH portion is a symlink that points to the currently active version of CDH and could be replaced with a more specific version folder, like `CDH-5.4.0-1.cdh5.4.0.p0.1/`.

rdblue: Looks like my last comment on this didn't display correctly, which is where the error came from. Sorry about that!

Install-Kite.md
DennisDawson: The indentation renders correctly in Chrome and Safari. Trying to continue the "c" indentation after the numbered indentation did not render properly when indented equal with "a" and "b". I had to play with this a bit to get it display properly. I looked for alternatives in the Markdown/Kramdown documentation, and couldn't find another solution.

Install-Kite.md
DennisDawson: The reason for these revisions is that the version of CDH is almost always going to have an older version of the kite-dataset command. These are the instructions for installing and verifying that the latest version is in use. I'm not referring to any specific version, because that would have to be updated with each release. I added the "current" link up here in response to an earlier comment.

Install-Kite.md
DennisDawson: added the spaces between the arguments

Install-Kite.md
DennisDawson: This doesn't make sense to me in this context. These instructions are for verifying that the version installed is the version in use.

PR #78, opened 2015-02-24
Remove old overview, update sidebar.

DennisDawson:


rdblue:
This looks fine, but I'm pretty sure there are more links to the data module overview that you're removing than just the sidebar. There's at least another one in the front page that needs to be updated as well.

DennisDawson:
That's the only instance I could find after a quick survey of the docs. If you find another, please file a JIRA issue. I'll update the sidebar and add a TOC after 1.0, once more documents are merged.

tomwhite:
I fixed a couple of redirects that redirected to the old page, and merged this. Thanks @DennisDawson!

PR #79, opened 2015-03-04
Include use of CLI flume-config.

DennisDawson:


DennisDawson:
On the 5.2 VM (for sure), the version of kite-dataset is "0.15.0-cdh5.2.0." That version doesn't include flume-config, which I'm using in the example. In preparing-the-vm.md, I should point to whatever we say in Kite-Install.md. PR 76 (https://github.com/kite-sdk/kite-docs/pull/76) addresses this issue. I'll point to those instructions rather than duplicating them in preparing-the-vm.md.

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,88 @@
+---
+layout: page
+title: Preparing the Virtual Machine
+---
+
+Complete the following steps to run Kite example code on a Cloudera Quickstart VM.
+
+* Install a VirtualBox or VMWare [Cloudera QuickStart VM][getvm] version 5.2 or later.
```
rdblue: Should note that VirtualBox is recommended. People generally have more problems with VMWare.

tutorials/generate-events.md
```diff
@@ -0,0 +1,69 @@
+---
+layout: page
+title: Generating Events
+---
```
rdblue: I think this tutorial needs more context:

* What are the prerequisites for running the steps? What previous tutorials does this assume I have completed?
* What does this accomplish?
* What will I learn from this tutorial?

tutorials/generate-events.md
```diff
@@ -0,0 +1,69 @@
...
+        .setUserId(userId)
+        .setSessionId(sessionId)
+        .setIp(ip)
+        .setTimestamp(randomTimestamp())
+        .setEventDetails(randomEventDetails())
+        .build();
+  }
+```
+
+## Running GenerateEvents
```
rdblue: This answers some of my questions above, so it should come first.

tutorials/generate-events.md
```diff
@@ -0,0 +1,69 @@
...
+
+Much of the class is devoted to creating random values. The two methods of interest are `run` and `generateRandomEvent`.
+
+The `run` method performs the following tasks:
+* creates a view of the `hive:events` dataset
+* creates a writer instance
+* spends 36 seconds writing random events
+* closes the writer, which stores the results in the `events` dataset.
+
+While the goal is to create random events, if they're _too_ random there won't be anything to aggregate. The `while` loop simulates a user session with random values for `sessionId`, `userId`, and `ip`. It then generates up to 25 random events for that session.
```
rdblue: I don't think people need to understand the code for events generation unless they are interested. Could you move the relevant information about _what_ is happening to the intro, then just cover _how_ in this section?

tutorials/generate-events.md
```diff
@@ -0,0 +1,69 @@
...
+  }
+```
+
+## Running GenerateEvents
+
+This example assumes that you've already created the [`hive:events` dataset][events].
+
+These are the steps to run the GenerateEvents program to populate the `hive:events` dataset. 
+
+1. In a terminal window, navigate to `/kite-examples/dataset`.
```
rdblue: These instructions should note that a prerequisite is cloning the examples repository and link back to the instructions.

tutorials/generate-events.md
```diff
@@ -0,0 +1,69 @@
...
+1. Enter `mvn compile`.
+1. Run the Java utility with `mvn exec:java -Dexec.mainClass="org.kitesdk.examples.data.GenerateEvents"`.
+
+Use Hue to view the records in Hive.
```
rdblue: Could you explain this a little more?

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,88 @@
...
+* In that VM, run the following command from a terminal window. This command clones a local copy of Kite code examples you can build and run.
+
+```bash
+git clone https://github.com/kite-sdk/kite-examples.git
+```
+
+* If you haven't already done so, download the [`kite-dataset` CLI JAR][install-cli].
```
rdblue: It would be better to use "command" to describe the CLI rather than referring to it as the"CLI JAR" or "executable JAR". The format doesn't matter much, what matters is that it is a command you can run directly. That's what you mean in nearly every case.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
+---
+layout: page
+title: Creating the Events Dataset
+---
+
+This lesson shows you how to create a dataset suitable for storing standard event records. You define a dataset schema, a partition strategy, and a URI that specifies the storage scheme.
```
rdblue: This should describe what a standard event record is.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
+---
+layout: page
+title: Creating the Events Dataset
```
rdblue: I think this tutorial needs more context as well. Maybe there should be an overall tutorial page that outlines the example goal, or maybe it can be included at the start of the tutorials. Otherwise, it isn't clear why we're creating "the" events dataset or what this is used for.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+    {
+      "name": "timestamp",
+      "type": "long",
+      "doc": "The point in time when the event occurred, represented as the number of milliseconds since January 1, 1970, 00:00:00 GMT. Required."
+    }
+  ]
+}
+```
+
+For convenience, save `standard_event.avsc` to the same directory where you installed the kite-dataset executable JAR.
```
rdblue: It would be better not to specify where this should be saved, as long as it is accessible. The user might have installed kite-dataset in their parcel directory so that location isn't always appropriate.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Defining the Partition Strategy
+
+Analytics for the `events` dataset are time-based. Partitioning the dataset on the `timestamp` field allows Kite to go directly to the files for a particular day, ignoring files outside the chosen time period. Partition strategies are defined in JSON format. See [Partition Strategy JSON Format][partition-strategies].
+
+The following code sample defines a strategy that partitions a dataset by _year_, _month_, and _day_, based on a _timestamp_ field.
```
rdblue: I wouldn't call this a code sample, it is a configuration.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+
+[partition-strategies]:{{site.baseurl}}/Partition-Strategy-Format.html
+
+## Creating the Events Dataset Using the Kite CLI
+
+Create the _events_ dataset using the default Hive scheme.
+
+To create the _events_ dataset:
+
+1. Open a terminal window and navigate to the directory where you saved the schema file.
```
rdblue: It would be easier to tell the user to go to a specific directory at the start instead of pointing them to the right one each time.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+  "type" : "month",
+  "name" : "month"
+}, {
+  "source" : "timestamp",
+  "type" : "day",
+  "name" : "day"
+} ]
+```
+
+For convenience, save `standard_event.json` to the same directory where you installed the `kite-dataset` executable JAR.
```
rdblue: This should be named according to what it does rather than after the schema it could be used in conjunction with.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+             --partition-by standard_event.json
+```
+
+Use Hue to look at the schema and confirm that the dataset is ready to use.
+
+[http://quickstart.cloudera:8888/filebrowser/view//tmp/data/default/events/.metadata/schema.avsc](http://quickstart.cloudera:8888/filebrowser/view//tmp/data/default/events/.metadata/schema.avsc)
+
+## Next Steps
+
+You've created a dataset to store events captured as they happen. Now you can run a web application to create records in your new dataset. See [Capturing Events with Flume][capture-events].
```
rdblue: I think the context given in the first sentence should be in the setup for this tutorial. The tutorial series creates a dataset to load with flume as events come in, but we actually use it in a few ways: we bulk-load fake data, stream events in through flume, and read events with Crunch. Saying that the dataset was created to do one of those is a little misleading and leaves the me wondering which part makes this dataset specific to storing events as they happen.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+
+```
+kite-dataset create events \
+             --schema standard_event.avsc \
+             --partition-by standard_event.json
+```
+
+Use Hue to look at the schema and confirm that the dataset is ready to use.
+
+[http://quickstart.cloudera:8888/filebrowser/view//tmp/data/default/events/.metadata/schema.avsc](http://quickstart.cloudera:8888/filebrowser/view//tmp/data/default/events/.metadata/schema.avsc)
```
rdblue: It isn't clear that this link is how the reader would use Hue to look at the schema. Plus, the dataset is in Hive so the reader should use the metastore browser instead of the file browser.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
+---
+layout: page
+title: Capturing Events with Flume
+---
+
+Once you have an [Events dataset][events], you can create a web application that captures session events.
+
+This example shows how you can send log information via Flume to your Hadoop database using a JSP and custom servlet running on Tomcat.
```
rdblue: This is great context, but I'd like to have a bit more information. At least:

* That you can use the generate-events tutorial as an alternative step if you want to move on and don't care about Flume.
* That these events will be sessionized in the next tutorial. This is to give a little more _why_ context.
* That the JSP and custom servlet running on Tomcat are not important. Those components are in place of the user's real application that has interesting events to look at and sessionize.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
...
+
+[events]:{{site.baseurl}}/tutorials/create-events-dataset.html
+
+## Configuring Flume
+
+These are the steps to configure Flume to channel log information directly to the `events` dataset. You first generate the configuration information using the Kite command-line interface, copy the results, paste them in the Flume configuration file, and then restart Flume.
+
+1. In a terminal window, type `kite-dataset flume-config --channel-type memory events`.
+1. Copy the output from the terminal window.
+1. Open Cloudera Manager.
```
rdblue: The VM setup instructions don't assume you're using Cloudera Manager, so I don't think this should either. The alternative is to copy the flume config and restart:

```
sudo cp flume.conf /etc/flume-ng/conf/flume.conf
sudo /etc/init.d/flume-ng-agent restart
```

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
...
+    pw.println("</body></html>");
+  }
+}
+```
+
+## Running the Web Application
+
+Follow these steps to build the web application, start the Tomcat server, and then use the web application to generate events that are sent to the Hadoop dataset.
+
+1. In a terminal window, navigate to `/kite-examples/demo`.
```
rdblue: Paths shouldn't start with "/" unless they are actually root paths. I think it would be better to have the user go to a specific directory and use relative paths in the tutorial.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
...
+1. Type the command `mvn install`.
+1. In the terminal window, enter `mvn tomcat7:run`.
+1. In a web browser, enter the URL [`http://quickstart.cloudera:8034/demo-logging-webapp/`][logging-app].
+1. On the web form, enter any user ID and a message, and then click **Send** to create a web event. 
+
+The Flume agent receives the events over inter-process communication (IPC), and the agent writes the events to the Hive file sink. Each time you send a message, Log4j writes a new `INFO` line in the terminal window.
```
rdblue: I think this should have more discussion and appear at the top of this tutorial. Readers are going to have the most questions about this part and the underlying architecture.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
...
+1. Choose the __Configuration__ tab.
+1. Click __Agent Base Group__.
+1. Right-click the Configuration File text area and choose __Select All__.
+1. Right-click the Configuration File text area and choose __Paste__.
+1. Click __Save Changes__.
+1. From the __Actions__ menu, choose __Restart__, and confirm the action.
+
+Flume is now configured to receive logging events and record them in the `events` dataset.
+
+## Creating Web Application Pages
```
rdblue: This has a lot of discussion on the servlet and JSP, but the focus should be on the architecture: The servlet (or any application) uses Log4j to "log" events, which are Avro records. Log4j is configured to send those events to Flume. Flume accumulates events and writes them to the Dataset.

Most people want to know  _why_ we are doing it this way:

1. Why Flume? Because it is reliable, widely used for this purpose, and has good integration.
2. Why not write directly to the dataset in Hadoop? Because we want to accumulate events to write into the same file without disrupting the web application. Flume will take responsibility for individual events and see that they are written in groups to the dataset. The application wants to move on after a single event.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,179 @@
...
+      pw.println("<p>Message: " + message + "</p>");
+      
+```
+
+Create a new StandardEvent builder.
+
+```Java
+      StandardEvent event = StandardEvent.newBuilder()
+```
+The event initiator is a user on the server. The event is a web message. These can be set as string literals, because the event initiator and event name are always the same.
```
rdblue: I think this might be a bug. Isn't the user a client and not a server?

rdblue:
Overall feedback:
* I think there generally needs to be more context to tell the reader why he needs to read the page, what he will learn, and what the next steps are. There should be a clear overall picture of what the whole tutorial series does and where each article fits.
* What the event generator does should be last and is a side discussion. The tutorial should focus on getting interesting events into the dataset and what interesting events are. The generator code isn't in the critical path to understanding the tutorial series, it is simply a helper so the reader has data to work with.
* The servlet and JSP shouldn't be the focus of the Flume pipeline tutorial. Flume, the configuration, what is happening, and why it is built that way should be the focus of that tutorial.

rdblue:
You might find the labs we built recently helpful for the background information. A lot of the Flume pipeline explanation is the same, though the specific example isn't and it doesn't use Log4j to send records.

  http://kitesdk.org/docs/0.17.1/labs/6-create-a-flume-pipeline.html

tutorials/preparing-the-vm.md
DennisDawson: If we say we recommend one over the other, even if it's true, it calls into question the quality of our VM and ruffles the feathers of a partner. It's fine to say that if one doesn't work, try the other, which covers both bases.

rdblue: Good point!

tutorials/generate-events.md
DennisDawson: I think you're right about this. I will add a section at the beginning of each tutorial concisely setting the context.

tutorials/generate-events.md
DennisDawson: What would be nice is if we had something interesting that exercised Kite in a practical way. I will handle the reorganization. Whatever code we provide as a sample needs to be described, at least in broad strokes.

tutorials/generate-events.md
DennisDawson: I'll deal with this in the context section at the start of the page. I don't want to continually link off the page throughout. At some point, the consumer needs to understand that these examples are modular and build on one another.

tutorials/generate-events.md
DennisDawson: I want to give the consumer some credit. If they're at a level where they are manipulating data using the API, they should be able to figure out how to view the records in Hive without a hand-holding lesson.

tutorials/preparing-the-vm.md
DennisDawson: I can't find an example where anyone says "download a command" when they mean "download a JAR." I don't know that consumers will be familiar with that terminology, and it would be confusing for them. I will check through, though, to ensure that I'm only discussing JARs in the context of download and install, but not for usage.

tutorials/create-events-dataset.md
DennisDawson: The next section goes into detail about what a standard event is. If the consumer is momentarily intrigued and wonders what a standard event could be, their suspense will be short-lived.

tutorials/create-events-dataset.md
DennisDawson: I'll add context at the start of each lesson.

tutorials/create-events-dataset.md
DennisDawson: I'll provide syntax that is "path to blah, path to blah" and give an example as if the source files are in the home directory.

tutorials/create-events-dataset.md
DennisDawson: Fine, but this is splitting hairs. JSON is a lightweight data interchange format based on a subset of JavaScript. Whatever we call it, it's a simple form of code. This is not confusing to the consumer.

tutorials/create-events-dataset.md
DennisDawson: It doesn't hurt to be sure they're in the correct directory when the steps are given in separate task sets.

tutorials/generate-events.md
rdblue: > Whatever code we provide as a sample needs to be described, at least in broad strokes.

Agreed, but I want for users to be able to understand what the tutorial accomplishes (add interesting data for later tutorials) and get it done before diving into it because it is mostly an interesting tangent. We should also note that it is optional.

> What would be nice is if we had something interesting that exercised Kite in a practical way

Could you explain what you mean here?

tutorials/preparing-the-vm.md
rdblue: > I can't find an example where anyone says "download a command" when they mean "download a JAR."

Fair enough. The real problem is that it is needlessly specific. What happens if we change the install so that it is no longer a JAR? Is there any value to calling it a JAR here? Fix it however you think works best.

tutorials/create-events-dataset.md
DennisDawson: I see that you now use `email-part.json` in one example, but most just say `user_part.json` or summat. I can follow this paradigm going forward, but it would be good to go back through existing examples and update for consistency.

tutorials/create-events-dataset.md
DennisDawson: I've discussed with Cris Morris ways that we can add context, and will address this concern with a new section at the top of each page.

tutorials/create-events-dataset.md
rdblue: Agreed. Please file a JIRA issue and assign yourself.

tutorials/create-events-dataset.md
DennisDawson: Yup - I'll change this to "use Hue Databrowsers-->Metastore Tables to examine the schema" It gives me a 500 error, but I'm sure that's a misconfiguration on my machine.

tutorials/flume-capture-events.md
DennisDawson: I'll add a context section to the top of each tutorial.

tutorials/flume-capture-events.md
DennisDawson: I'll add a context section to the top of each tutorial.

tutorials/flume-capture-events.md
DennisDawson: Adding context will help, but the essential issue here is that all we're doing is configuring Flume to watch for events, and that's achieved using the CLI. There's very little Kite stuff involved in this example.

tutorials/create-events-dataset.md
rdblue: Actually, I think it does. This makes the user keep track of a decision he or she made earlier, rather than being clear about what they should do.

tutorials/flume-capture-events.md
DennisDawson: I agree that the event generator is orthogonal to the discussion. It would be helpful if https://github.com/kite-sdk/kite-examples/pull/23 were addressed, to ensure that the code is sound, or at least not embarrassing.

tutorials/flume-capture-events.md
rdblue: > There's very little Kite stuff involved in this example.

I agree, but that highlights the need to make what is happening very clear. Focusing on the web application doesn't  help the user understand the message we want them to, but discussing why they care about Kite's support in Flume does.

tutorials/flume-capture-events.md
rdblue: Is that ready for review? I thought you were having trouble with command-line options. I'm happy to review it when you think it's ready.

DennisDawson: It is ready for review. It runs, and does what I want it to do.

tutorials/preparing-the-vm.md
DennisDawson: Changed it to "install the Kite CLI."

tutorials/create-events-dataset.md
DennisDawson: Moot.

tutorials/flume-capture-events.md
DennisDawson: I added a separate list of instructions for configuring Flume from the command line.

tutorials/flume-capture-events.md
DennisDawson: Changed to "user on the client."

tutorials/flume-capture-events.md
DennisDawson: Removed the /.

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,93 @@
...
+     ```
+
+1. If you haven't already done so, install the [Kite CLI][install-cli].
+
+[getvm]: http://www.cloudera.com/content/support/en/downloads/quickstart_vms.html
+[install-cli]:{{site.baseurl}}/Install-Kite.html
+
+## Configuring the VM
+
+Some Kite examples require Flume. If you use Cloudera Manager, Flume user impersonation is configured for you. If do not use Cloudera Manager, you must update Flume user impersonation in `core-site.xml`.
```
rdblue: I think we need to add more background on Cloudera Manager. The steps above get you a working VM with Kite, but don't mention CM, while this assumes the user knows about the choice and has already decided whether to use CM. I think a small section to introduce that choice is needed. It should include the fact that we recommend using CM, how to start it (opening the link on the desktop, I believe), and why you would choose not to use it (it requires more memory allocated to the VM).

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,93 @@
...
+---
+## Purpose
+This lesson describes the steps for configuring a virtual machine to run Kite example code on a Cloudera Quickstart VM.
+
+### Result
+Your VM is ready for you to install and run sample programs from the Kite SDK Examples project.
+
+## Installing the VM and Kite Examples
+
+1. Install a VirtualBox or VMWare [Cloudera QuickStart VM][getvm] version 5.2 or later.
```
rdblue: I think this should be "Download and open" or "Download and import" the VM.

With the note about CM, this should warn users to increase memory if they intend to use CM before starting the appliance.

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,93 @@
+---
+layout: page
+title: Preparing the Virtual Machine
+---
+## Purpose
+This lesson describes the steps for configuring a virtual machine to run Kite example code on a Cloudera Quickstart VM.
+
+### Result
```
rdblue: The prerequisite for this lab is that the user has either VMWare or VirtualBox installed. Links to download those tools would be helpful.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
...
+---
+## Purpose
+
+This lesson shows you how to create a dataset suitable for storing standard event records, as defined in [The Unified Logging Infrastructure for Data Analytics at Twitter][paper]. You define a dataset schema, a partition strategy, and a URI that specifies the storage scheme.
+
+[paper]:http://vldb.org/pvldb/vol5/p1771_georgelee_vldb2012.pdf
+
+### Prerequisites
+
+A VM or cluster with CDH installed.
```
rdblue: I think this prerequisite should be that the reader has completed the VM setup steps (with a link). We don't intend for this to run on a random CDH cluster or without having already checked out the example repository and installed `kite-dataset`.

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,93 @@
...
+  1. Find the _Virtualization_ settings (usually under _Security_), and enable all virtualization options.
+
+* __How do I get my mouse back?__
+  * If your mouse/keyboard is stuck in the VM (captured), you can usually release it by pressing the right `CTRL` key. If you don't have one (or if that didn't work), click the release key in the lower-right corner of the VirtualBox window.
+
+* __Other problems__
+  * Using VirtualBox? Try using VMWare.
+  * Using VMWare? Try using VirtualBox.
+
+[vbox]: https://www.virtualbox.org/wiki/Downloads
```
rdblue: Please add a section for context that links to the next tutorial(s) in the series.

tutorials/preparing-the-vm.md
```diff
@@ -0,0 +1,93 @@
+---
+layout: page
+title: Preparing the Virtual Machine
+---
+## Purpose
+This lesson describes the steps for configuring a virtual machine to run Kite example code on a Cloudera Quickstart VM.
+
+### Result
+Your VM is ready for you to install and run sample programs from the Kite SDK Examples project.
```
rdblue: I'd like this to be more specific. This is what happens, but the reason why we want these steps completed is to move on to the next tutorial. There should be a link here as well as a link towards the bottom.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
+---
+layout: page
+title: Creating the Events Dataset
+---
+## Purpose
+
+This lesson shows you how to create a dataset suitable for storing standard event records, as defined in [The Unified Logging Infrastructure for Data Analytics at Twitter][paper]. You define a dataset schema, a partition strategy, and a URI that specifies the storage scheme.
```
rdblue: I think this should differentiate between the skills that are demonstrated and the purpose in the context of the tutorials. The skill is to use a given Avro schema and partition strategy to create a dataset. The context for the tutorials is that this module has the reader create a dataset for the standard event records that are used by subsequent modules. The next modules should also be linked somewhere in the context section.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
...
+
+[paper]:http://vldb.org/pvldb/vol5/p1771_georgelee_vldb2012.pdf
+
+### Prerequisites
+
+A VM or cluster with CDH installed.
+
+### Result
+
+You create `dataset:hive:events`, where you can store standard event objects. You can use the dataset with several Kite tutorials that demonstrate data capture, storage, and analysis.
```
rdblue: I like "You can use the dataset with...". Could you add links to those tutorials?

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
...
+      "type": "long",
+      "doc": "The point in time when the event occurred, represented as the number of milliseconds since January 1, 1970, 00:00:00 GMT. Required."
+    }
+  ]
+}
+```
+
+## Defining the Partition Strategy
+
+Analytics for the `events` dataset are time-based. Partitioning the dataset on the `timestamp` field allows Kite to go directly to the files for a particular day, ignoring files outside the time period. Partition strategies are defined in JSON format. See [Partition Strategy JSON Format][partition-strategies].
```
rdblue: Can you define what you mean by analytics are time-based? An example would help understanding.

rdblue: What are the assumptions for users reading this tutorial? This paragraph assumes the reader is familiar with partitioning, which may not be the case. I think it would be better to explain it in this order:
1. Typical uses are time-limited: you care about a day or an hour of the data, not all of it.
2. To take advantage of this and not read data you don't need, use a `year/month/day` organization for the data.
3. This is called _partitioning_ and configured with a _partition strategy_. Include links to relevant reference docs.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
...
+
+[partition-strategies]:{{site.baseurl}}/Partition-Strategy-Format.html
+
+## Creating the Events Dataset Using the Kite CLI
+
+Create the _events_ dataset using the default Hive scheme.
+
+To create the _events_ dataset:
+
+1. Open a terminal window.
```
rdblue: Nit: This doesn't need to be a step because we don't want the user to open a new terminal each time. You could instead say "In a terminal window, run the `create` command . . ." That would get rid of the need for numbering here since it's just one command and some explanation.

tutorials/create-events-dataset.md
```diff
@@ -0,0 +1,109 @@
...
+1. Use the `create` command to create the dataset. This example assumes that you stored the schema and partition definitions in your home directory. Substitute the correct path if you stored them in a different location.
+
+```
+kite-dataset create events \
+             --schema ~/standard_event.avsc \
+             --partition-by ~/partition_year_month_day.json
```
rdblue: I think it would be more clear to have the user do this in a known directory, like the kite-examples directory. This would remove the need to discuss directories other than to include a step to go to the appropriate one at the start of each tutorial.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Configuring Flume in Cloudera Manager
+
+1. In a terminal window, type `kite-dataset flume-config --channel-type memory events`.
+1. Copy the output from the terminal window.
+1. Open Cloudera Manager.
+1. Under __Status__, click the link to __Flume__.
+1. Choose the __Configuration__ tab.
+1. Click __Agent Base Group__.
+1. Right-click the Configuration File text area and choose __Select All__.
```
rdblue: Nit: "Configuration File" should be distinguished as something to look for with formatting.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Configuring Flume
+
+Follow these steps to configure Flume to channel log information directly to the `events` dataset. You first generate the configuration information using the Kite command-line interface, copy the results, paste them in the Flume configuration file, and then restart Flume.
+
+You can configure Flume for this example using either Cloudera Manager or the command line.
+
+### Configuring Flume in Cloudera Manager
+
+1. In a terminal window, type `kite-dataset flume-config --channel-type memory events`.
```
rdblue: I think the command could use more explanation. What doe sit do and why? What might I change for production instead of the demo? (The channel would be a file channel in production, by the way)

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+1. Click __Save Changes__.
+1. From the __Actions__ menu, choose __Restart__, and confirm the action.
+
+### Configuring Flume from the Command Line
+
+1. In a terminal window, enter `kite-dataset flume-config --channel-type memory events -o flume.conf`.
+1. To update Flume configuration, enter `sudo cp flume.conf /etc/flume-ng/conf/flume.conf`.
+1. To restart the Flume agent, enter `sudo /etc/init.d/flume-ng-agent restart`.
+
+Flume is now configured to listen for web application events and record them in the `events` dataset.
```
rdblue: This doesn't need to state "web application" events. Just events. It would be great to explain more of the config and what this is doing with Flume. As it is, this treats Flume as a black box and doesn't help the reader understand what is going on inside it. Kite's configuration is a starting point, so it is important for the reader to know what Kite is telling Flume to do.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+1. To compile the application, enter `mvn install`.
+1. To start the Tomcat server, enter `mvn tomcat7:run`.
+1. In a web browser, enter the URL [`http://quickstart.cloudera:8034/demo-logging-webapp/`][logging-app].
+1. On the web form, enter any user ID and a message, and then click **Send** to create a web event. 
+
+View the log messages in the terminal window where you launched Tomcat. View the records in Hive using the Hue File Browser.
+
+[logging-app]:http://quickstart.cloudera:8034/demo-logging-webapp/
+
+## Creating Web Application Pages
```
rdblue: I think this should be "Understanding" not "Creating" because the app is already written.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
+---
+layout: page
+title: Capturing Events with Flume
+---
+
+## Purpose
+
+This lesson demonstrates how you can configure Flume to capture events from a web application with minimal impact on performance or the user. Flume collects individual events and writes them in groups to the dataset.
+
+The Flume agent receives the events over inter-process communication (IPC), and writes the events to the Hive file sink. Each time you send a message, Log4j writes a new `INFO` line in the terminal window.
```
rdblue: I don't think this needs to mention IPC. IPC covers a wide variety of communication methods. What actually happens is the Flume agent listens for events that are sent from an application, in this case the sample web application. That application sends events by "logging" them through Log4j, which also logs the event to the terminal.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+This example demonstrates how to generate Flume configuration information from the Kite CLI. In addition, JSP and servlet samples allow you to test the data capture mechanism.
```
rdblue: The web app demonstrates how to send events to Flume using Log4j. You can use the web app to send a few sample events. Is that what you meant by "test the data capture mechanism"?

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+* A VM or cluster configured with Flume user impersonation. See [Preparing the Virtual Machine][vm].
+* An [Events dataset][events] in which to capture session events.
+
+[vm]:{{site.baseurl}}/tutorials/preparing-the-vm.html
+[events]:{{site.baseurl}}/tutorials/create-events-dataset.html
+
+### Result
+
+Flume is configured to listen for events on a Tomcat server instance. Use the JSP and servlets to send events to Tomcat. Log4j logs each event to the terminal window. Flume stores the events in `dataset:hive:events`.
```
rdblue: Tomcat is just a service that runs the web application. Flume doesn't work with Tomcat at all, though the application running in Tomcat is sending events to Flume.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Configuring Flume
+
+Follow these steps to configure Flume to channel log information directly to the `events` dataset. You first generate the configuration information using the Kite command-line interface, copy the results, paste them in the Flume configuration file, and then restart Flume.
```
rdblue: I think this section should explain what the Kite command generates and what Flume is doing as a result.

tutorials/flume-capture-events.md
rdblue: I noted above where the extra explanation should go.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. In a terminal window, enter `kite-dataset flume-config --channel-type memory events -o flume.conf`.
+1. To update Flume configuration, enter `sudo cp flume.conf /etc/flume-ng/conf/flume.conf`.
+1. To restart the Flume agent, enter `sudo /etc/init.d/flume-ng-agent restart`.
+
+Flume is now configured to listen for web application events and record them in the `events` dataset.
+
+## Running the Web Application
+
+Follow these steps to build the web application, start the Tomcat server, and then use the web application to generate events that are sent to the Hadoop dataset.
```
rdblue: This section should start with context about how the web application sends data to Flume. That's the most important part of it and the other half of the important tutorial content.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+```Java
+          .build();
+      logger.info(event);
+    }
+    pw.println("<p><a href=\"/demo-logging-webapp\">Home</a></p>");
+    pw.println("</body></html>");
+  }
+}
+```
```
rdblue: There should be some summary, next steps, etc. here.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
...
+
+The application obtains the session ID and IP address from the request object, and creates a timestamp based on the local machine clock.
+
+```Java
+          .setSessionId(request.getSession(true).getId())
+          .setIp(request.getRemoteAddr())
+          .setTimestamp(System.currentTimeMillis())
+```
+
+Build the StandardEvent object, and then send the object to the logger with the level _info_.
```
rdblue: This is the important part. Log4j is configured to pass these events to Flume, so this is where the data is actually sent from the application.

tutorials/flume-capture-events.md
rdblue: I think the logging configuration and StandardEvent code should be explained above this point, so the rest of the tutorial is optional for those readers that want to understand the rest of the web app in detail.

tutorials/generate-events.md
```diff
@@ -0,0 +1,84 @@
+---
+layout: page
+title: Generating Events
+---
+## Purpose
+
+Kite applications work with Big Data. This example class, `GenerateEvents.java`, generates 1-1.5 million random event records, a small amount of realistic Big Data you can use with Kite examples.
```
rdblue: The purpose is to fill the dataset with sample data that can be used in the following tutorials. This simulates the data that a web application (link to the Flume web app tutorial) would send if it were running for a long time. Because we don't have a long time for it to run, this is needed.

I don't think you need the part about "Big Data" because it doesn't clarify what is happening and this doesn't produce a big dataset. In the end, this introduces "Big Data" to say that this tutorial module doesn't produce big data.

tutorials/flume-capture-events.md
```diff
@@ -0,0 +1,199 @@
+---
+layout: page
+title: Capturing Events with Flume
+---
+
+## Purpose
```
rdblue: Somewhere in this section, please note that this is optional and the reader can generate data with the other tutorial. That way, readers that aren't interested in Flume can skip this.

tutorials/generate-events.md
```diff
@@ -0,0 +1,84 @@
...
+1. Creates a view of the `hive:events` dataset.
+1. Creates a writer instance.
+1. Spends 36 seconds writing random events.
+1. Closes the writer, which stores the results in the `events` dataset.
+
+Although the goal is to create random events, if they're _too_ random, there won't be anything to aggregate. The `while` loop simulates a user session with random values for `sessionId`, `userId`, and `ip`. It then generates up to 25 random events for that session.
+
+```Java
+    View<StandardEvent> events = Datasets.load(
+        (args[0].isEmpty() ? "dataset:hive:events" : args[0]),
```
rdblue: I don't recommend using this pattern here. It is much better to use a variable called `datasetUri`.

tutorials/generate-events.md
```diff
@@ -0,0 +1,84 @@
...
+Much of the class GenerateEvents creates random values. The two methods of interest are `run` and `generateRandomEvent`.
+
+The `run` method performs the following tasks:
+
+1. Creates a view of the `hive:events` dataset.
+1. Creates a writer instance.
+1. Spends 36 seconds writing random events.
+1. Closes the writer, which stores the results in the `events` dataset.
+
+Although the goal is to create random events, if they're _too_ random, there won't be anything to aggregate. The `while` loop simulates a user session with random values for `sessionId`, `userId`, and `ip`. It then generates up to 25 random events for that session.
```
rdblue: I don't think the idea of a session has been introduced anywhere. This should link back to the module that explains what this tutorial accomplishes -- aggregating individual web requests into user sessions. I'm not sure that exists anywhere.

tutorials/preparing-the-vm.md
DennisDawson: VM installation is in the instructions, with a link to the download location.

tutorials/preparing-the-vm.md
DennisDawson: These examples are modular and do not necessarily lead from one to the next. The progression is shown in a table of contents, not in the modules themselves.

tutorials/preparing-the-vm.md
DennisDawson: This is handled in the TOC, so as not to make individual modules out of date as others are added or removed.

tutorials/generate-events.md
DennisDawson: The idea of a session is introduced in this paragraph. It simulates a session, and generates 25 random events.

tutorials/generate-events.md
DennisDawson: This pattern allows the user to run the application without arguments, as in the current published example, but also allows the user to choose a different dataset. If you're saying that I should use the same pattern to set the variable before this statement, that adds another step, but I don't see how it changes the outcome. Why is it "much better" to use a variable?

tutorials/preparing-the-vm.md
DennisDawson: Install is fine.

My original draft had the configuration for CM, but I was told to remove it. I can add it back in.

tutorials/preparing-the-vm.md
DennisDawson: I'll add a section on CM, with the 8GB memory and 2 CPU recommended configurations. This is documented elsewhere and should not be a barrier to publication of our examples, even if we choose to include it.

tutorials/create-events-dataset.md
DennisDawson: The modules stand alone. The structure is provided by a TOC. The skills description is fine.

tutorials/create-events-dataset.md
DennisDawson: There's no reason this shouldn't run on any CDH cluster. I agree that they should have already installed kite-dataset.

tutorials/create-events-dataset.md
DennisDawson: The modules stand alone. Structure is provided by a TOC.

tutorials/create-events-dataset.md
DennisDawson: Added links in the Purpose section, to cover conceptual topics.

tutorials/create-events-dataset.md
DennisDawson: I prefer this as is.

tutorials/create-events-dataset.md
DennisDawson: We've been back and forth on the directory. The home directory is fine for now. I might revisit this in the future.

tutorials/flume-capture-events.md
DennisDawson: Readers who are not interested in Flume will not go to a page titled "Capturing Events with Flume."

PR #80, opened 2015-03-12
CDK-754: Kite Data API: Kite and Crunch

DennisDawson:


tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
+---
+layout: page
+title: Aggregating Records with Crunch
+---
+
+## Purpose
+
+The Apache Crunch Java library provides a framework for writing, testing, and running applications in MapReduce pipelines.
+
+This tutorial demonstrates how you use Kite's `CrunchDatasets` helper methods to configure source and target datasets for Crunch operations that aggregate individual events into work sessions.
```
rdblue: As I noted on the other pull request, this should refer back to the section that describes the overall setup for this series of tutorials. I think that context is needed for the reader to know what sessions you are talking about.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+
+Create the `sessions` dataset using the default Hive scheme.
+
+To create the sessions dataset:
+
+1. Open a terminal window.
+1. Use the `create` command to create the dataset. This example assumes that the schema file is in your home directory. Change the path to `session.avsc` if you stored it in another directory.
+
+```
+kite-dataset create sessions --schema ~/session.avsc
```
rdblue: This can assume the user hasn't changed directories, so there's no need for the `~/` part of the path or discussion about location.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+Use Hue to confirm `sessions` is in your list of Hive tables.
```
rdblue: Other sections link to Hue, which would be useful here.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+
+```Java
+    PCollection<StandardEvent> events = read(
+        CrunchDatasets.asSource(eventsDataset));
+```
+
+Step 2. Process the data.
+a. Create a session key by combining the user ID and session ID.
+b. Group together all events with the same session key.
+c. For each group, create a Session record as an Avro Specific object.
```
rdblue: I think the bit about the Avro Specific object is distracting. The main idea is to create a session record. If you think it is valuable to point out that it is a specific object (which I would agree with) then it can be done later in discussion.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+
+```Java
+    getPipeline().write(sessions,
+        CrunchDatasets.asTarget("dataset:hive:sessions"),
+        Target.WriteMode.APPEND);
+    return run().succeeded() ? 0 : 1;
+  }
+```
+
+The `GetSessionKey` class combines the session ID with the user ID to create a unique session key.
```
rdblue: I think this should note that this is the implementation of the step used above.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+
+Inspect each event in this session group. Track the earliest timestamp (start time) and latest timestamp (end time). Keep a count of the events in this session group.
+
+```Java
+      while (events.hasNext()) {
+        final StandardEvent event = events.next();
+        
+```
+
+Reset the start time if the timestamp is earlier than the current start time value.
```
rdblue: I don't think it is accurate to say the value is "reset".

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+Keep a count of the events in this session group.
+
+```Java
+        numEvents += 1;
+      }
+```
+
+Create a session using the Crunch `emitter`. Use values from the first event in the group for fields that don't change. Crunch repeats the `emit` process for each group of events until all sessions are processed.
```
rdblue: The emitter isn't used to create the session. That's the job of the Session builder, which is created with `Session.newBuilder()`. The emitter is used to send the session to the next phase of processing.

Crunch doesn't repeat emit, it keeps running this method for each group of events. This method emits a single session for each group.

tutorials/crunch-aggregate-sessions.md
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  public static void main(String... args) throws Exception {
+    int rc = ToolRunner.run(new AggregateEvents(), args);
+    System.exit(rc);
+  }
+
+}
+
+```
```
rdblue: This should include more explanation at the end.

PR #81, opened 2015-04-27
Add documentation on "dataFormat" partitioning strategy.

HelmutZechmann:
Just complemented missing docs on the dateFormat partitioning strategy.


rdblue:
Thanks @HelmutZechmann!

I have to admit though, we were holding off on this to avoid encouraging people to use it. Right now, we don't have the same level of support for dateFormat that we do for year/month/day/etc. The issue to update it is [CDK-960|https://issues.cloudera.org/browse/CDK-960]. I highly recommend using the separate partitioners for now.

